PR #12, opened 2015-01-13
added missing autotools files

pfasante:
adds the files requested for in https://github.com/dstehle/fplll/issues/11

### DIFF
--- a/src/bkz.cpp
+++ b/src/bkz.cpp
@@ -93,9 +93,9 @@ bool BKZReduction<FT>::svpReduction(int kappa, int blockSize, const BKZParam &pa
     double cputimeStart2 = cputime();
 
     for(int i=0; ; i++) {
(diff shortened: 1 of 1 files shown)


malb:
I think this is bad practice as it will create tons and tons of merge conflicts and will only produce frustration. If you want an easier way for people to get started, you shoul write an ./autogen.sh bash script which runs autoreconf -i and creates the m4 directory.

pfasante:
ok, I agree with that, did not had it in mind. Perhaps there should be a note in the readme about autoreconf then?

malb:
Yep, that'd be good. In fact, someone (hint hint) should convert the README to markdown so it's displayed on github appropriately and update it to explain how to build the library when checking out of git.

pfasante:
Good, I'll take a look at it

PR #14, opened 2015-01-15
Added options for GH enum bound in BKZ and own enum bound in SVPPruning

jvdpol:
This adds the BKZ_GH_BND option to the BKZ flags and the ghFactor parameter to the BKZParam class. When BKZ_GH_BND is set, the enumeration bound maxDist is set to ghFactor*GH^2, where GH is the norm of the expected shortest vector in the block by the Gaussian Heuristic. If no vector is found when BKZ_GH_BND is set, the old vector is kept instead. Note that this will no longer guarantee the final basis to be BKZ-reduced.

Since I used the exponential function to compute GH, I required some hack to patch the dpe exponential function which is not actually implemented. I tried to emulate the patch for the dpe log function (which I think computes log base 10 for some reason).

### DIFF
--- a/src/fplll.cpp
+++ b/src/fplll.cpp
@@ -27,8 +27,6 @@
 
 FPLLL_BEGIN_NAMESPACE
 
-THIS IS BROKEN, DEAL WITH IT
-
 template<class ZT>
 void zerosFirst(ZZ_mat<ZT>& b, ZZ_mat<ZT>& u, ZZ_mat<ZT>& uInvT) {
   int i, d = b.getRows();


src/svpcvp.cpp
```diff
@@ -157,7 +157,9 @@ static int shortestVectorEx(IntMatrix& b, IntVect& solCoord,
     FPLLL_CHECK(result, "shortestVector: cannot compute an initial bound");
     maxDist.add(maxDist, ftmp1, GMP_RNDU);
   }
-
+  
+  if(flags & SVP_OVERRIDE_BND) maxDist.set_z(argIntMaxDist, GMP_RNDU);
```
malb: I think there should be an if/else block as it's easier to understand it as an alternative. Currently, this simply overwrites a computation done earlier and it's not clear to the reader if this is a potential bug or by design.

src/defs.h
```diff
@@ -206,7 +207,8 @@ enum BKZFlags {
   BKZ_MAX_TIME = 8,
   BKZ_BOUNDED_LLL = 0x10,
   BKZ_AUTO_ABORT = 0x20,
-  BKZ_DUMP_GSO = 0x40
+  BKZ_DUMP_GSO = 0x40,
+  BKZ_GH_BND = 0x80
```
malb: Would it make sense to add a command line option to main.ccp so it's easier to test?

malb:
Hi Joop, sorry for the late reply. I have a few more suggestions, I hope they are not unreasonable: I added some comments to the diff. Additionally, could you rebase your commit to the current master? Also, did you test it. Is it faster (for some inputs)? Can you give an example where it's faster.

PR #22, opened 2015-01-28
BKZ with GH bound + SVPPruning with user bound

jvdpol:
Rebasing screwed up my commit chain so I have recreated this pull request. 

It adds the possibility of using the Gaussian Heuristic enumeration bound in BKZ with the BKZ_GH_BND flag and the ghFactor parameter or the -bkzghbound command with a ghFactor. BKZ will then use ghFactor*GH^2 as its enumeration bound instead of the usual (length of the current shortest vector), but only if the block size of the enumeration is bigger than 30. If a vector is not found during the enumeration, the current vector is kept and it continues with the next index. Note that this is a heuristic, and BKZ-reducedness will no longer be guaranteed. In order to compute the GH value it uses the exponential function which was not properly implemented for dpe, so I added a hack fix to dpe.h (and nr.cpp/nr.h), similar to how the log function was "fixed" for dpe.

Note that this heuristic will mostly be useful when used in conjunction with pruning on high block sizes, but here is some experimental result: On an 80-dimensional GM lattice it takes about 46 seconds to do 5 rounds of BKZ without the heuristic and 36 seconds with the heuristic on my machine. The output GSO is slightly worse with the heuristic due to missed vectors. Again, though, the effect will be much more noticeable with higher BKZ block size, but it is not as feasible to test this without pruning. 

For SVP it adds an SVP_OVERRIDE_BND flag, which can be used with the shortestVectorPruning function to choose your own bound which does not get the added error term of other SVP calls. This allows you to implement some extreme pruning enumeration with your own bound. 

Additionally, I fixed a bug where bkzlinearpruning was not initialized and one where the bkzmaxloops value was not passed on to BKZ, rendering the command useless.

### DIFF
--- a/src/bkz.cpp
+++ b/src/bkz.cpp
@@ -72,6 +72,32 @@ static double getCurrentSlope(MatGSO<Integer, FT>& m, int startRow, int stopRow)
   return v1 / v2;
 }
 
+template<class FT>
+static void computeGaussHeurDist(MatGSO<Integer, FT>& m, FT& maxDist, long maxDistExpo, int kappa, int blockSize, double ghFactor){
+  double t = (double)blockSize/2.0+1;
+  t=tgamma(t);
+  t=pow(t,2.0/(double)blockSize);
+  t=t/M_PI;
+  FT f,g,h;
+  f = t;
+  m.getR(h,kappa,kappa);
+  g.log(h);
+  for(int i = kappa+1; i < kappa+blockSize; i++){
+      m.getR(h,i,i);
+      h.log(h);
+      g.add(g,h);
+  }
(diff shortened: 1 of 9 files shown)


malb:
I'm happy with this pull request and suggest that it's merged.

I tried to find an example yesterday, though, where it would give a nice speed up but in my unscientific attempts I didn't manage to find any. Anyway, it seems good to have it as an option.

jvdpol:
Doing a full 60-dimensional enumeration on a Goldstein Mayer lattice with GH takes 14h23m and without GH it takes 17h18m which is about 20% more. It's not going to make larger block size in BKZ any more feasible than it is now, though.

PR #25, opened 2015-03-14
LLL Simple Implementation C++

prabhdeep311:


prabhdeep311:
I hardcoded the dimensions just to whip up a quick implementation. Will fix that and the index issue.
Floats were used because earlier I was using GS normalization, but now I'll switch to int. Or would it be better using Matrix<ZT> ?

malb:
You should definitely use templates here, just to get comfortable with using them.

src/lll-implement.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+    int i = 0;
+    for(int j = 0; j < ineachbasis; j++) b[i][j] = a[i][j];
+    
+    B[0] = dotProduct(b, 0, b, 0, ineachbasis);
+    
+    
+    for(i = 1; i<noofbasis; i++)
+    {
+        for(int j = 0; j < ineachbasis; j++) b[i][j] = a[i][j];
```
malb: Is there no copyRow() method?

src/lll-implement.cpp
```diff
@@ -0,0 +1,115 @@
...
+    int dot= 0;
+    for(int k = 0; k < ineachbasis; k++)
+    {
+        dot+=(a[i][k]*b[j][k]);
+    }
+    return dot;
+}
+
+//Apply Gram-Schmidt Process, matrix b gives the Gram-Schmidt basis
+void GSO(Matrix<int>& a, Matrix<int>& b, int noofbasis, int ineachbasis, int B[], Matrix<float>& u)
```
malb: Matrices know their dimensions, why pass them?

malb: It is convention to have "return values" such as B and u at the front. Also, mark inputs which are not modified by const. Programming is about communicating with other people too, not only about "communicating" with computers.

src/lll-implement.cpp
```diff
@@ -0,0 +1,115 @@
...
+                RED(a, u, k, j, noofbasis);
+        }
+        if(B[k] >= (0.75 - (u[k][k-1]*u[k][k-1]))*B[k-1] )	//Lovasz Condition (delta) = 0.75
+        {
+            k++;
+        }
+        else	//Lovasz Condition fails, swap rows and calculate new values
+        {
+            for(int p = 0; p < ineachbasis; p++) swap(a[k][p], a[k-1][p]);
+            GSO(a, b, noofbasis, ineachbasis, B, u);
```
malb: Do yo need to recompute all of the Gram-Schmidt matrix here?

PR #26, opened 2015-03-15
Generating random numbers for Z_NR

mihir-w:
Currently, random number generation works for Z_NR class only when the data type is mpz_t.
Have added methods to generate random numbers for types long and double using random number functions of GMP.

### DIFF
--- a/src/nr.cpp
+++ b/src/nr.cpp
@@ -210,6 +210,27 @@ inline long Z_NR<long>::exponent() const {
 }
 
 template<>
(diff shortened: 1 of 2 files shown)


shi-bai:
This seems good to me. Thanks Mihir and Martin. I merged this request.

PS: the function randm() now only generates non-negative random integers. But this is a inherited-feature from mpz random functions. Either we could force the fplll randm() to produce only non-negative random integers; we can further tune the function randm() to follow the sign from its parameter max.

PR #28, opened 2015-03-22
Structure to test suite.

mihir-w:
Changes done:

1. Created a folder named tests and wrote 2 simple tests in it for integer multiplication and checking of lll conditions respectively.
2. Added simple error messages using FPLLL_CHECK.
3. Added above two files in TESTS variable in Makefile.am. `make check` command now runs these 2 tests.
4. Shifted testing commands from Makefile.am into test_general script and ran the script from the Makefile.am

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -6,8 +6,10 @@ include_fplll_HEADERS=defs.h dpe.h fplll.h matrix.h matrix.cpp nr.h \
 EXTRA_DIST=example_in example_out example_svp_in example_svp_out dim55_in \
 	pruning.txt
 
+TESTS = simple_intmul simple_lll
+
 bin_PROGRAMS=fplll latticegen
-check_PROGRAMS=llldiff
+check_PROGRAMS=$(TESTS) llldiff
 EXTRA_PROGRAMS=fplll_dbg latticegen_dbg test 
 lib_LTLIBRARIES=libfplll.la
 EXTRA_LTLIBRARIES=libfplllv.la libfpllld.la
@@ -28,6 +30,12 @@ latticegen_dbg_LDADD=libfpllld.la
 
 llldiff_SOURCES=llldiff.cpp
 llldiff_LDADD=libfplll.la
+
+simple_intmul_SOURCES = tests/simple_intmul.cpp
+simple_intmul_LDADD = libfplll.la
+
+simple_lll_SOURCES = tests/simple_lll.cpp
+simple_lll_LDADD = libfplll.la
 
 test_SOURCES=test.cpp
 test_LDADD=libfplll.la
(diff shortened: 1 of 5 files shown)


malb:
This is a good start! Thanks! Note that the automatic test suite fails, which in a way confirms that your code works :) Could you fix that? Also, it would be nice if your test programs would indicate where they failed to stdout.

mihir-w:
Oh! I was generating random numbers upto 2^30 which caused overflow and hence the failure.

> Also, it would be nice if your test programs would indicate where they failed to stdout.

I have rewritten simple_intmul.cpp in [catch](https://github.com/philsquared/Catch) framework. All it required was to put catch.hpp in tests folder and make some simple changes in simple_intmul.cpp

Here's the failure log I got: [failure_log](http://pastebin.com/gxy6HqzW)
After correction: [success_log](http://pastebin.com/HhyDjSi9)
Both logs are color-coded. I could not get the color scheme on pastebin.

Also since I put catch.hpp in tests folder, it's showing +9000 insertions because of that. How can I give the proper credit ?

malb:
Hi,

this looks good. 

1) It is fine to just check in catch.hpp. The file has its own copyright information, so you're not taking false credit for checking it in. To keep things clean, you might want to add it in its own commit, though.

2) If I may be nitpicky: could you adapt the coding style (spaces, newlines) to that of fplll? It is good to be consistent here for readability.

mihir-w:
I have changed the coding style as much as I can. Do let me know if I missed something.
Also I have rewritten simple_lll.cpp using catch.

In simple_lll, sometimes the gram-schimdt coefficients are slightly more than 0.5 after lll reduction. As a result test_sizeReduced fails in such cases.
Example: Cell(8, 3): 5.08144e-1 [failed_lll](http://pastebin.com/iz6kBEi6)
I guess it would be better to change the comparison to allow some leeway ?

malb:
Yep, some numerical leeway will be required.

malb:
Okay, I finally got around to clone your pull request locally. Catch looks like a nice enough testing framework/

Two small comments:

- tests shouldn't be a subdirectory of src but should be on the same level as src
- tests should get its own Makefile.am (currently automake will complain because of the subdir)

I know automake and friends can be daunting and annoying, but in the long run doing things properly is going to be less work.

mihir-w:
> tests shouldn't be a subdirectory of src but should be on the same level as src

I am planning to have 2 test files for nearly every header file. One would cover simple unit tests and the other file would cover high-level tests ( Properties of algorithms. Time-consuming). This way user can run only the fast unit tests when he just wants to check some minor changes or check installation.
Would'nt keeping them in same src folder cause too much distraction ?  Or perhaps you mean that the tests folder should be in fplll/ instead of fplll/src

> tests should get its own Makefile.am (currently automake will complain because of the subdir)

Yes, I agree. I was thinking on the same line. The placement of test files would resolve this too.

malb:
Indeed, I simply meant to move fplll/src/tests to fplll/tests, sorry for not being clearer.

shi-bai:
Thanks Mihir. The current test framework seems good for me. I agree with Martin that they should be put into a folder fplll/tests. 

For the LLL/BKZ test, what I had in mind was that: there should be a mechanism to make sure a newer version of FPLLL will do LLL/BKZ with no-worse quality for the same inputs (with default parameters). The "same inputs" can be generated by using seeds in latticegen.cpp. The correctness and the quality of the reduced basis  in LLL/BKZ can be checked and compared as per the method in your discussion with Martin. Maybe the first step is to check the correctness, say for svp, check that the returned lattice point is indeed in the lattice as mentioned by Martin.

PR #30, opened 2015-03-27
Adds documentation to bkz.h.

TimonKnigge:
Documents some functions in the header-file for the bkz-algorithm (for computing the shortest vector).


malb:
looks good

PR #39, opened 2015-11-20
dual enumeration

blowfish880:
This adds the ability to enumerate shortest vectors in the dual instead of the primal to the Enumeration class. For details on the dual enumeration algorithm, see http://eprint.iacr.org/2015/1123. 

Access to the new enumeration is provided through the shortestVector function (in svpcvp.cpp) and a new flag: SVP_DUAL.

### DIFF
--- a/src/svpcvp.cpp
+++ b/src/svpcvp.cpp
@@ -55,9 +55,10 @@ static void getBasisMin(Integer& basisMin, const IntMatrix& b,
 static bool enumerateSVP(int d, MatGSO<Integer, Float>& gso, Float& maxDist,
         Evaluator<Float>& evaluator, const vector<double>& pruning,
         int flags) {
-  if (d == 1 || !pruning.empty()) {
+  bool dual = (flags & SVP_DUAL);
+  if (d == 1 || !pruning.empty() || dual) {
     Enumeration::enumerate(gso, maxDist, 0, evaluator, FloatVect(), FloatVect(),
-            0, d, pruning);
+            0, d, pruning, dual);
   }
   else {
     Enumerator enumerator(d, gso.getMuMatrix(), gso.getRMatrix());
@@ -121,18 +122,27 @@ static int shortestVectorEx(IntMatrix& b, IntVect& solCoord,
     //FPLLL_TRACE("Ignoring the last " << d - newD << " vector(s)");
     d = newD;
   }
-
-  if (evalMode == EVALMODE_SV) {
-    /* Computes a bound for the enumeration. This bound would work for an
-       exact algorithm, but we will increase it later to ensure that the fp
-       algorithm finds a solution */
-    getBasisMin(intMaxDist, b, 0, d);
+  
+  if (flags & SVP_DUAL) {
+    maxDist = gso.getRExp(d - 1, d - 1);
+    Float one; one = 1.0;
+    maxDist.div(one, maxDist);
+    if (flags & SVP_VERBOSE) {
+      cout << "maxDist = " << maxDist << endl;
+    }
+  } else {
+    if (evalMode == EVALMODE_SV) {
+      /* Computes a bound for the enumeration. This bound would work for an
+         exact algorithm, but we will increase it later to ensure that the fp
+         algorithm finds a solution */
+      getBasisMin(intMaxDist, b, 0, d);
+    }
+    else {
+      /* Use the bound given as a parameter */
+      intMaxDist = argIntMaxDist;
+    }
+    maxDist.set_z(intMaxDist, GMP_RNDU);
   }
-  else {
-    /* Use the bound given as a parameter */
-    intMaxDist = argIntMaxDist;
-  }
-  maxDist.set_z(intMaxDist, GMP_RNDU);
 
   // Initializes the evaluator of solutions
   Evaluator<Float>* evaluator;
@@ -155,7 +165,11 @@ static int shortestVectorEx(IntMatrix& b, IntVect& solCoord,
     Float ftmp1;
     bool result = evaluator->getMaxErrorAux(maxDist, true, ftmp1);
     FPLLL_CHECK(result, "shortestVector: cannot compute an initial bound");
-    maxDist.add(maxDist, ftmp1, GMP_RNDU);
+    if (flags & SVP_DUAL) {
+      maxDist.add(maxDist, ftmp1, GMP_RNDU);
+    } else {
+      maxDist.sub(maxDist, ftmp1, GMP_RNDU);
+    }
   }
 
   // Main loop of the enumeration
--- a/src/svpcvp.cpp
+++ b/src/svpcvp.cpp
@@ -165,11 +165,7 @@ static int shortestVectorEx(IntMatrix& b, IntVect& solCoord,
     Float ftmp1;
     bool result = evaluator->getMaxErrorAux(maxDist, true, ftmp1);
     FPLLL_CHECK(result, "shortestVector: cannot compute an initial bound");
-    if (flags & SVP_DUAL) {
-      maxDist.add(maxDist, ftmp1, GMP_RNDU);
-    } else {
-      maxDist.sub(maxDist, ftmp1, GMP_RNDU);
-    }
+    maxDist.add(maxDist, ftmp1, GMP_RNDU);
   }
   
   // Main loop of the enumeration
--- a/src/bkz.cpp
+++ b/src/bkz.cpp
@@ -93,9 +93,9 @@ bool BKZReduction<FT>::svpReduction(int kappa, int blockSize, const BKZParam &pa
     double cputimeStart2 = cputime();
 
     for(int i=0; ; i++) {
-      if ((par.flags & BKZ_MAX_LOOPS) && i >= par.maxLoops) break;
-      if ((par.flags & BKZ_MAX_TIME) && (cputime() - cputimeStart2) * 0.001 >= par.maxTime) break;
-      if (autoAbort.testAbort(par.autoAbort_scale, par.autoAbort_maxNoDec)) break;
+      if ((preproc->flags & BKZ_MAX_LOOPS) && i >= preproc->maxLoops) break;
+      if ((preproc->flags & BKZ_MAX_TIME) && (cputime() - cputimeStart2) * 0.001 >= preproc->maxTime) break;
+      if (autoAbort.testAbort(preproc->autoAbort_scale, preproc->autoAbort_maxNoDec)) break;
 
       bool clean2 = true;
       if (!bkzLoop(i, dummyKappaMax, *preproc, kappa, kappa + blockSize, clean2))
--- a/src/defs.h
+++ b/src/defs.h
@@ -190,7 +190,8 @@ enum SVPMethod {
 
 enum SVPFlags {
   SVP_DEFAULT = 0,
-  SVP_VERBOSE = 1
+  SVP_VERBOSE = 1,
+  SVP_DUAL = 2
 };
 
 enum CVPFlags {
--- a/src/enumerate.cpp
+++ b/src/enumerate.cpp
@@ -1,4 +1,5 @@
-/* Copyright (C) 2008-2011 Xavier Pujol.
+/* Copyright (C) 2008-2011 Xavier Pujol
+   (C) 2015 Michael Walter.
 
    This file is part of fplll. fplll is free software: you
    can redistribute it and/or modify it under the terms of the GNU Lesser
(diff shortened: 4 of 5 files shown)


blowfish880:
I did not. Since in the primal case the code is almost identical to before, the only difference being this one check, I guess I just assumed that it would not affect it too much. But I guess since it's in a loop, it might be better to pull this check out and have 2 separate loops when computing the centerPartSums. But then again, I guess the hope would be that the compiler can do something like this itself. Not sure, might be worth checking.

malb:
To test the performance impact, I ran the following test (using fpylll) with and without this pull request

```python
import time
from fpylll import *
    
n = 44
    
for i in range(10):
    A = IntegerMatrix(n, n)
    A.randomize("ntrulike", bits=32, q=1456817003)
    print A[0, -1],
    LLL.reduction(A)
    
    M = GSO.Mat(A)
    M.update_gso()
    max_dist, expo = M.get_r_exp(0, 0)
    
    t = time.clock()
    Enumeration.enumerate(M, max_dist, expo, 0, n, None)
    print time.clock() - t
```

The output on my machine is

<table>
<thead>
<tr>
<th scope="col" class="org-right">instance</th>
<th scope="col" class="org-right">with #39</th>
<th scope="col" class="org-right">w/o #39</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-right">883302933</td>
<td class="org-right">5.662741</td>
<td class="org-right">4.490525</td>
</tr>


<tr>
<td class="org-right">822154706</td>
<td class="org-right">6.935338</td>
<td class="org-right">5.780412</td>
</tr>


<tr>
<td class="org-right">2308615747</td>
<td class="org-right">8.359915</td>
<td class="org-right">6.731231</td>
</tr>


<tr>
<td class="org-right">4024222343</td>
<td class="org-right">22.99513</td>
<td class="org-right">18.363614</td>
</tr>


<tr>
<td class="org-right">315063919</td>
<td class="org-right">8.218641</td>
<td class="org-right">6.484413</td>
</tr>


<tr>
<td class="org-right">472743699</td>
<td class="org-right">1.858108</td>
<td class="org-right">1.47106</td>
</tr>


<tr>
<td class="org-right">4049222344</td>
<td class="org-right">23.353564</td>
<td class="org-right">18.4365</td>
</tr>


<tr>
<td class="org-right">55830938</td>
<td class="org-right">0.00037</td>
<td class="org-right">0.000307</td>
</tr>


<tr>
<td class="org-right">644730440</td>
<td class="org-right">52.821682</td>
<td class="org-right">41.004684</td>
</tr>


<tr>
<td class="org-right">747529060</td>
<td class="org-right">0.00272</td>
<td class="org-right">0.002121</td>
</tr>
</tbody>
</table>

The timings vary a bit from run to run, but overall those timings seem stable. 

Unfortunately, it seems this PR as it stands slows down enumeration noticeably, plausibly because of the introduction of many conditional jumps. I didn't investigate further, though.

shi-bai:
Thanks Michael for the code (and Martin for the test). It may take some
time for me read your paper 2015/1123 <http://eprint.iacr.org/2015/1123> and
absorb the code. I hope you don't mind.

Best,
Shi

blowfish880:
Thank you Martin, for looking into this. I was quite surprised to see such a large gap (note to self: be more careful with this kind of stuff), so I looked a little into it. As it turns out, this gap is to a large part due to the counting of explored nodes that I introduced for benchmarking and left in there. This is done using Z_NR<mpz_t> structure to avoid overflow, but as it turns out, this hurts performance quite a bit. When I removed it, things sped up, but there is still a gap, which seem to be due to the conditional jump. Pulling the check out of the for loop and using two separate loops (as I suggested as a possibility above), decreases the gap even further, but a small one remains. This last one can be virtually removed by duplicating the entire "enumerateLoop" function - one for the primal and one for the dual loop. However, this involves a lot of code duplication and I am not entirely sure how desirable this is. What do you guys think?

malb:
Thanks Michael. I guess we'll have to check how small the difference turns out to be. One other option to try could be to use GCC's (and Clangs?) __builtin_expect, like here

https://github.com/malb/m4ri/blob/3be7c636ff3ac8ef47bf0a794c6e8fb439b83d51/src/misc.h#L443

I'm definitely in the camp that tries to avoid code duplication. I'm happy to help & try options.

PS: counting explored nodes is definitely something we should offer as an option.

blowfish880:
I did a few more tests to get some concrete numbers. I ran all the options I mentioned above (and the base line test) for 10 identical lattice bases in dimension 45. This is what the slow down looks like:
For my original implementation (minus node counting, minimal code duplication), the slow down is between 7-8%. 
When pulling the check out of the for loop (~3 lines of code duplication), the difference is between 3-4%.
When duplicating the "enumerateLoop" (~50 lines of code duplication), the difference is between 1-1.5%. 

I have not looked into the __builtin_expect thing. 

I have removed the node counting so far for the sake of comparison, but I think we can probably include it efficiently using "long long" - enumerating more than 2^64 nodes seems unrealistic. 

opinions?

malb:
I'd say 1.-1.5% should be within measurement errors, so that would be okay in my book. 3-4% is not ideal but perhaps a price worth paying for not duplicating 50 lines of code. Shall we do the node counting in a separate pull request?

src/svpcvp.cpp
```diff
@@ -155,7 +166,11 @@ static int shortestVectorEx(IntMatrix& b, IntVect& solCoord,
     Float ftmp1;
     bool result = evaluator->getMaxErrorAux(maxDist, true, ftmp1);
     FPLLL_CHECK(result, "shortestVector: cannot compute an initial bound");
-    maxDist.add(maxDist, ftmp1, GMP_RNDU);
+    if (flags & SVP_DUAL) {
+      maxDist.add(maxDist, ftmp1, GMP_RNDU);
```
shi-bai: Is it maxDist.sub here?

blowfish880: Actually, now that I think about it, I believe this is a relic from when I inverted the maxDist bound for dual enumeration later. Now I think the original line here would be the correct thing to do. Thank you for pointing that out. I'll update the pull request soon, where I'll include the changes from above as well.

blowfish880:
I updated the pull request. I think node counting should be a separate pull request, as it is a different feature. I also briefly looked at __builtin_expect, which allows one to give hints for branch prediction to the compiler. Unfortunately, I don't think that is very useful in our context. We would maybe be able to speed up primal enumeration with this, but this would come at the cost of slowing down enumeration, right? What we would want ideally, that the compile generates 2 separate functions - one with dual=false and one with dual=true and removes the conditional jumps. Any idea on how to do that? 
Other than that, how does it look now? Can you confirm that it is faster now than before?

malb:
The C way of doing these conditional compilations would be 

```
#define PRIMAL 1
#include <actual_implementation.h>
```

in some `primal.cpp`  and 

```
#define PRIMAL 0
#include <actual_implementation.h>
```

in some `dual.cpp`. Then `actual_implementation.h` would have code

```
if (PRIMAL) {
   foo
}
```

which is then optimised away by the compiler. Maybe the proper C++ way is to use templates (?)

blowfish880:
Interesting. This looks like it would do the trick. I guess the "proper" C++ way might be to use inheritance and virtual functions. I hope I'll find time over the holidays to try these things.

pfasante:
Wouldn't virtual function implementations be dispatched during runtime - and thus would be slower than if dispatched during compilation time (which would be the case in templated functions)?

blowfish880:
Generally, this is true. But from what I've read, under certain conditions virtual functions can be inlined. I'm not entirely sure what these conditions are and if they would hold in such an implementation in our case, but I think it might be worth checking, since it seems more natural to me than templates. I think, using templates for this is not the natural use case for them and might result in somewhat awkward code. But I'm far from being an expert on this and am certainly open to being educated otherwise.

blowfish880:
I just pushed a new version. This one can live without any code duplication, only has minor modifications so the original code, and has a slow down of only 2% in my experiments. It's the result of the realization that through some simple reordering of the data and the introduction of a pointer, one can keep the enumeration code uniform. The remaining 2% gap is due to the pointer having to be dereferenced a lot (multiple times per explored node).

malb:
I'm seeing a slowdown of about 5% here (same Python code as above):

   Instance |  with #39 | without #39 | ratio 
------------ | ------------- | ------------- |
 2770147048 |  5.835726 |    5.565045 | 1.048 
 1412562793 | 18.808205 |   17.925677 | 1.049 
 3438523787 |  4.066015 |    3.873364 | 1.049 
 3128283224 |  3.89e-05 |     4.0e-05 | 0.972 
 3691354213 | 19.860664 |   19.017978 | 1.044 
 2140286795 |  2.249763 |     2.14503 | 1.048 
  971621894 |  3.489196 |    3.328356 | 1.048 
 3836136518 |  0.098274 |    0.093751 | 1.048 
 1478693123 |  1.714205 |    1.633752 | 1.049 
 1472913774 |  6.366569 |     6.06239 | 1.050 

I'll also try a different machine.

malb:
I ran 

```python
import time
from fpylll import *
from copy import copy

n = 44

set_random_seed(0)

for i in range(10):
    A = IntegerMatrix(n, n)
    A.randomize("ntrulike", bits=32, q=1456817003)
    print "%10d"%A[0, -1],
    LLL.reduction(A)

    T = []
    for j in range(5):
        B = copy(A)
        M = GSO.Mat(B)
        M.update_gso()
        max_dist, expo = M.get_r_exp(0, 0)

        t = time.clock()
        Enumeration.enumerate(M, max_dist, expo, 0, n, None)
        T.append(time.clock() - t)
    print "%.3f"%min(T),
    print "%.3f"%max(T)
```

on my laptop and got:

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">
<colgroup>
<col  class="org-right" />
<col  class="org-right" />
<col  class="org-right" />
<col  class="org-right" />
<col  class="org-right" />
<col  class="org-right" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-right">Instance</th>
<th scope="col" class="org-right">with #39 min</th>
<th scope="col" class="org-right">with #39 max</th>
<th scope="col" class="org-right">w/o #39 min</th>
<th scope="col" class="org-right">w/o #39 max</th>
<th scope="col" class="org-right">min/min</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-right">2770147048</td>
<td class="org-right">5.873</td>
<td class="org-right">6.102</td>
<td class="org-right">5.698</td>
<td class="org-right">5.744</td>
<td class="org-right">1.030</td>
</tr>


<tr>
<td class="org-right">1412562793</td>
<td class="org-right">18.913</td>
<td class="org-right">19.421</td>
<td class="org-right">18.410</td>
<td class="org-right">19.100</td>
<td class="org-right">1.027</td>
</tr>


<tr>
<td class="org-right">3438523787</td>
<td class="org-right">4.065</td>
<td class="org-right">4.196</td>
<td class="org-right">4.048</td>
<td class="org-right">4.469</td>
<td class="org-right">1.004</td>
</tr>


<tr>
<td class="org-right">3128283224</td>
<td class="org-right">0.000</td>
<td class="org-right">0.000</td>
<td class="org-right">0.000</td>
<td class="org-right">0.000</td>
<td class="org-right">1.000</td>
</tr>


<tr>
<td class="org-right">3691354213</td>
<td class="org-right">20.330</td>
<td class="org-right">20.624</td>
<td class="org-right">19.872</td>
<td class="org-right">20.319</td>
<td class="org-right">1.023</td>
</tr>


<tr>
<td class="org-right">2140286795</td>
<td class="org-right">2.301</td>
<td class="org-right">2.312</td>
<td class="org-right">2.244</td>
<td class="org-right">2.267</td>
<td class="org-right">1.025</td>
</tr>


<tr>
<td class="org-right">971621894</td>
<td class="org-right">3.570</td>
<td class="org-right">3.593</td>
<td class="org-right">3.510</td>
<td class="org-right">3.532</td>
<td class="org-right">1.017</td>
</tr>


<tr>
<td class="org-right">3836136518</td>
<td class="org-right">0.100</td>
<td class="org-right">0.102</td>
<td class="org-right">0.096</td>
<td class="org-right">0.101</td>
<td class="org-right">1.041</td>
</tr>


<tr>
<td class="org-right">1478693123</td>
<td class="org-right">1.749</td>
<td class="org-right">1.764</td>
<td class="org-right">1.717</td>
<td class="org-right">1.735</td>
<td class="org-right">1.018</td>
</tr>


<tr>
<td class="org-right">1472913774</td>
<td class="org-right">6.542</td>
<td class="org-right">6.619</td>
<td class="org-right">6.382</td>
<td class="org-right">6.450</td>
<td class="org-right">1.025</td>
</tr>
</tbody>
</table>

I added min/max to see if the spread within one case is much bigger than the difference between the two cases. Overall, the ration here is somewhere around 2-3%.

malb:
I guess we should discuss if we accept this slow down or not. I'd be okay with it. Of course, if it could be avoided, that'd be better.

malb:
How about we swallow the bitter pill and allow for ~50 lines of code duplication by having two `enumerateLoop`? This way, the slow down is minimal and everybody is happy :)

blowfish880:
Before doing the code duplication (which I really hate - I think it's the source of all evil in this world), I figured I'd give it another shot using templates. This is now essentially duplicating the `enumerateLoop` function so I get essentially the same running times as without the changes. Can you confirm this, Martin?

malb:
Yep, I can confirm it. Whoohoo, merge time! Thank you!

PR #41, opened 2015-11-23
Quaddouble Support

malb:
Support for double doubles and quad double.


shi-bai:
Hi Martin,

Thanks for merging the cmake->make and correct some errors brought by me :) The code seems good. I will then add a -dd/-qd option in main.cpp.

Cheers,
Shi

PR #43, opened 2015-11-30
updated readme

shi-bai:


malb:
Maybe the README should mention that libqd is optional?

shi-bai:
sure, updated a bit more.

PR #47, opened 2016-01-06
Minor tweak to get it building on cygwin.

m8pple:
All the long double templates are disabled for cygwin, but one template instantiation
in fplll.cpp wasn't guarded. Previously failed with:
  .libs/fplll.o:fplll.cpp(.text$_ZN5fplll6MatGSOINS_4Z_NRIA1_12__mpz_structEENS_5FP_NRIeEEE12updateGSORowEii[_ZN5fplll6MatGSOINS_4Z_NRIA1_12__mpz_structEENS_5FP_NRIeEEE12updateGSORowEii]+0xba): undefined reference to `fplll::FP_NR<long double>::is_nan() const'

and multiple other missing definitions for FP_NR<long double>.

I added an #ifdef guard to disable that one too.

I haven't tried the fix on any other platforms, only cygwin.


malb:
Thanks!

PR #54, opened 2016-01-30
bugfix in the dual enumeration

blowfish880:
After the dual enumeration completed, the maxDist norm was not updated correctly. So the function worked and returned a shortest dual vector, but the maxDist variable still contained the old value.


malb:
Ha, good example of "code that isn't tested is broken". We really should add tests for this stuff.

PR #91, opened 2016-06-22
add codes for tuple gauss sieve

shi-bai:


src/sieve/sampler_basic.cpp
```diff
@@ -0,0 +1,172 @@
...
+  for (int i = 0; i < nr; i++) {
+    tmp.set_z(g(i,i));
+    ((*s_prime)[i]).div(maxbistar2, tmp);
+    ((*s_prime)[i]).sqrt((*s_prime)[i], GMP_RNDN);
+  }
+
+  /* verbose */
+  set_verbose(ver);
+  print_param();
+  gmp_randinit_default(state);
```
malb: Why does the sampler maintain its own state? Wouldn't it be better to use the common state for reproducibility?

src/sieve/sampler_basic.h
```diff
@@ -0,0 +1,99 @@
...
+  private:
+
+  /**
+   * GSO object of B
+   */
+  int nr, nc;
+  double logn2;
+  gmp_randstate_t state;
+  ZZ_mat<ZT> b, u, uInv;
+  MatGSO<Z_NR<ZT>, F> *pGSO;
```
malb: My C++ isn't strong enough but couldn't this just be a MatGSO object (not a pointer)?

src/sieve/sampler_main.cpp
```diff
@@ -0,0 +1,108 @@
+#include "sampler_basic.h"
```
malb: Shouldn't this live in tests?

src/sieve/sieve_Gauss.cpp
```diff
@@ -0,0 +1,323 @@
+/*
```
malb: maybe sieve_gauss.cpp?

malb:
Also, it seems the sieving code isn't actually built.

src/sieve/sieve_Gauss.cpp
shi-bai: done, fixed above. It seems compilting now with
make latsieve
there was some change in the codebase I guess which made it broken :)

shi-bai:
delete this pull request as my fork has been removed previously..

PR #114, opened 2016-06-24
Including the pruner

lducas:
Adding a pruner to fplll.


fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+
+
+#include <array>
+#include "factorial.const"
+#include "ballvol.const"
+#include "fplll.h"
+
+FPLLL_BEGIN_NAMESPACE
+
+// This file propose an implementation of the numerically optimized
```
malb: proposes an implementation? Isn't it an implementation? :)

fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+// pruning strategy of [GNR10]. 
+
+// Many details for implemantation follows from the thesis [Chen13]
```
malb: typo: implementation

fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+// Some simplification have been made, for example we restrict ourselves
```
malb: simplifications

fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+    
+    Pruner();
+
+    template<class GSO_ZT,class GSO_FT>
+    void load_basis_shape(MatGSO<GSO_ZT, GSO_FT>& gso, int beginning = 0, int end = 0);
+    void load_basis_shape(int dim, double* gso_sq_norms);
+    
+    //void set_parameters(FT preproc_cost, FT target_success_proba);
+    
+    void optimize_pruning_coeffs(/*io*/double* pr, /*i*/ int reset = 1);
```
malb: Can some of these inputs be tagged const?

fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+    inline FT svp_success_proba(/*i*/ evec &b);
+    inline FT cost_factor(/*i*/ evec &b);
+    FT cost_factor_derivative(/*i*/ evec &b, /*o*/ evec &res);
+    int improve(/*io*/ evec &b);
+    void descent(/*io*/ evec &b);
+
+    FT tabulated_factorial[PRUNER_MAX_N];
+    FT tabulated_ball_vol[PRUNER_MAX_N];
+    
+    FT one;  // HACK: because we don't have (double - FT) (yet ?)
```
malb: Maybe open an issue for this so it doesn't get lost. Shouldn't be hard to fix.

fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+    FT shell_ratio; // Shell thickness Ratio when evaluating svp proba
+    FT symmetry_factor; // Set at 2 for SVP enumeration assuming the implem only explore half the space
+};
+
+template<class FT>
+Pruner<FT>::Pruner(){
+  n = 0;
+  d = 0;
+  set_tabulated_consts();
+  epsilon = std::pow(2., -13);
```
malb: Shall we add some text to explain how these constants were chosen?

fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+  target_success_proba = .90;
+  preproc_cost = 0;
+  min_cf_decrease = .9999;
+  symmetry_factor = 2;
+}
+
+
+template<class FT>
+void Pruner<FT>::set_tabulated_consts(){
+  mpfr_t tmp;
```
malb: Why not use FP_NR<mpfr_t> here?

fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+// naming conventions:
+// b is for bound (squared)
+// pv is for partial volumes (NOT squared)
+// r is for gram schmidt length (squared). Automatically renormalized
+// to avoid overflowing partial volumes
+// p is for polynomial
+
+// inside this code, b,pv,and R are in reversed order
+// as to conform with the algorithm desciption of [Chen13]
+// reversing output and ouput is done by the C extern function
```
malb: 'extern "C"' has a specific meaning in C++ (link using C convention, not C++ convention). This is not what you mean here, right?

fplll/test.cpp
```diff
@@ -19,17 +19,47 @@
 using namespace std;
 using namespace fplll;
 
+
```
malb: Is editing test.cpp a mistake? (I'd guess this file should be removed, in general)

fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+#define PRUNER_MAX_D 1023
+#define PRUNER_MAX_N 2047
+
+
+
+template<class FT> 
+class Pruner{
+  public:
+    class TestPruner;
+    friend class TestPruner;
```
malb: Mhh, I'm not necessarily such a big fan of a class having to know its test class. But I don't know what the best practices are in C++

fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+  FT dxn = pow_si(dx, 2 * d);
+  FT dvol =  dxn * relative_volume(d, b_minus_db) - vol;
+  return dvol / (dxn - 1.);
+
+}
+
+
+
+template<class FT>
+inline FT Pruner<FT>::cost_factor(/*i*/ evec &b){
```
malb: Is there a better name than cost_factor perhaps? It's not really a factor (?)

fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+    void init_prunning_coeffs(evec &b);
+    void load_prunning_coeffs(/*i*/double* pr, /*o*/ evec& b);
+    void save_prunning_coeffs(/*o*/double* pr, /*i*/ evec& b);
+    inline int enforce(/*io*/ evec &b, /*opt i*/ int j = 0);
+    inline FT eval_poly(int ld,/*i*/ poly *p, FT x);
+    inline void integrate_poly(int ld,/*io*/ poly *p);
+    inline FT relative_volume(/*i*/int rd, evec &b);
+    inline FT cost(/*i*/ evec &b);
+    inline FT svp_success_proba(/*i*/ evec &b);
+    inline FT cost_factor(/*i*/ evec &b);
```
malb: Maybe a one sentence summary/documentation for each function?

fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+
+
+/// Autoprune function, hiding the Pruner class
+
+
+template<class FT, class GSO_ZT, class GSO_FT> 
+void auto_prune(/*output*/ double* pr, double& success_proba,
+                /*inputs*/ double enumeration_radius, double preproc_cost,
+                double target_success_proba,
+                MatGSO<GSO_ZT, GSO_FT>& gso, int beginning = 0, int end = 0){
```
malb: The old pruner supported passing several matrix shapes to the pruner at once, is that option gone?

fplll/pruner.h
```diff
@@ -0,0 +1,568 @@
...
+
+
+
+
+
+/// Autoprune function, hiding the Pruner class
+
+
+template<class FT, class GSO_ZT, class GSO_FT> 
+void auto_prune(/*output*/ double* pr, double& success_proba,
```
malb: Shall we return the cost here as well?

fplll/pruner.h
lducas: Leftover comment from the standalone pruner.

fplll/pruner.h
lducas: The only alternative I know is making all methods public...

fplll/pruner.h
lducas: Done

fplll/pruner.h
lducas: Done.

fplll/pruner.h
lducas: It would be nice to have a read_from_string function in the FP_NR class indeed. Adding it as an issue.

fplll/pruner.h
lducas: I can surely find a simple way of re-adding this feature. Maybe for the next pull request ?

fplll/test.cpp
lducas: Removing it indeed.

fplll/pruner.h
lducas: Why not. Which of single_enum_cost or repeated_enum_cost would make more sense ?

fplll/test.cpp
lducas: Hum. Removing that file seems to have been a bad idea. make distr now
fails. I guess there is something to remove from a config file ?

2016-06-30 18:21 GMT+02:00 Martin Albrecht <notifications@github.com>:

malb: Lines 55 and 56 need to be dropped as well: https://github.com/fplll/fplll/blob/master/fplll/Makefile.am#L55

malb:
Closing this PR, because code was moved to different branch.

PR #122, opened 2016-07-01
Various fixes

malb:


malb:
@cr-marcstevens  can you look at my commit fea2b69 to check if it deals with files & strings the C++ way?

cr-marcstevens:
Hi Martin,

For C++ I would avoid C strings except for constants.
Note that C strings (constants) can be automatically converted to a std::string,
so functions that take 'const std::string&' can also take a C string,
but of course that doesn't work the other way around (passing std::string for a C string parameter).

Regarding files, in C++ we have std::ifstream.
Checking if a path can be opened for reading takes just a single line:

if ( std::ifstream(path).fail() )
OR
if ( std::ifstream(path).good() )

these statements:
- construct temporary std::ifstream
- tries to open the file
- returns a boolean on the outcome
- automatically destructs std::ifstream and closes file if it was opened


E.g.:
header:
const std::string& default_strategy_directory();

code:
const std::string& default_strategy_directory()
{
	// static (='global') constant string object constructed only once
	static const std::string ret(FPLLL_DEFAULT_STRATEGY_DIR);
	return ret;
}

const std::string strategy_full_path(const std::string& strategy_path)
{
	if ( std::ifstream(strategy_path).good() )
		return strategy_path;
	std::string path = default_strategy_directory() + "/" + strategy_path;
	if ( std::ifstream(path).good() )
		return path;
	path.clear();
	return path;
}

vector<Strategy> load_strategies_json(const std::string& filename)
{
	json js;
	{
		std::ifstream fs(filename);
  		fs >> js; // this is actually a more common C++ form than 'js << fs;'
	} // destructs fs
	..

PR #131, opened 2016-07-05
clean flag in svp reduction now based on length of first vector

blowfish880:
Also, call to LLL inside randomization loop and removed from end of rerandomization. 

Preprocessing and postprocessing still return clean flags, which are unused. Should they remain? if not, should they return something else?


fplll/bkz.cpp
```diff
@@ -158,10 +158,7 @@ template <class FT> void BKZReduction<FT>::rerandomize_block(int min_row, int ma
     }
   }
   m.rowOpEnd(min_row, max_row);
-
-  // 3. LLL reduce
-  if (!lll_obj.lll(0, min_row, max_row))
```
malb: Why remove this check? It checks for numerical instability, doesn't it?

fplll/bkz.cpp
```diff
@@ -310,16 +307,14 @@ bool BKZReduction<FT>::dsvp_postprocessing(int kappa, int block_size, const vect
...
-    throw std::runtime_error(RED_STATUS_STR[lll_obj.status]);
-  }
-
-  clean &= (lll_obj.nSwaps == 0);
+  
+  int first = dual ? kappa + block_size - 1 : kappa;
+  
+  // ensure we are computing something sensible
+  m.updateGSORow(0, first);
+  FT old_first;
```
malb: We often use getRExp in BKZ, I have no preference, but maybe we should do this here as well for consistency?

blowfish880: you're right. we absolutely should you getRExp (to avoid overflows). i changed it. in the process i also replaced the gso update call with a size_reduction, since again i was noticing some weird numerical behavior (which only affected the clean flag in non-critical situations though).

fplll/bkz.cpp
```diff
@@ -310,16 +307,17 @@ bool BKZReduction<FT>::dsvp_postprocessing(int kappa, int block_size, const vect
...
-  bool clean = true;
-
   int lll_start = (par.flags & BKZ_BOUNDED_LLL) ? kappa : 0;
-
-  if (!lll_obj.lll(lll_start, lll_start, kappa + block_size))
-  {
+  
+  int first = dual ? kappa + block_size - 1 : kappa;
+  
+  // ensure we are computing something sensible
```
malb: Maybe expand in the comment on when skipping size reduction here could go wrong? It' the kind of line people will be tempted to optimise away in the future :)

malb:
Have you checked the test failures? `test_bkz` segfaults. I can take a look if you like.

blowfish880:
i've been looking at it this morning. it seems like there is an issue with calling size_reduction without prior call to lll (the variables babaiMu and babaiExpo are extended to the required size by the lll function). it's strange though since it was working fine, before i did the last merge and i have no idea what affected this.

fplll/bkz.cpp
```diff
@@ -499,6 +493,18 @@ template <class FT> bool BKZReduction<FT>::bkz()
...
+  {
+    m.updateGSO();
+    sld_potential = m.get_slide_potential(0, num_rows, param.block_size);
+  }
+  
+  // the following is necessary, since sd-bkz starts with a dual tour and 
+  // svp_reduction calls size_reduction, which needs to be preceeded by a 
+  // call to lll lower blocks to avoid seg faults
+  if (sd)
+    lll_obj.lll(0, 0, num_rows);
```
malb: Don't we always have to either assume or enforce that the input is LLL reduced, i.e. shouldn't we make this initial LLL call unconditional?

blowfish880: usually this is handled by the wrapper in the bkz_reduction function in fplll.cpp (except if the BKZ_NO_LLL flag is set, in which case calling lll beforehand would be incorrect). so the input is already lll reduced. the issue is just that certain variables are not initialized in the lll_obj that we have at hand in the bkz object, so size_reduction can segfault if called on lower blocks right away (as we do in sd, but not in bkz or sld).

malb: I'm leaning towards unconditional calling regardless, just to ensure a sane state, but perhaps people really really want BKZ_NO_LLL so they shall have it. Merge time.

blowfish880: we could always call it in case BKZ_NO_LLL is not set.

malb: I've added a new issue to discuss this.

Issue #134, opened 2016-07-06
Buggy pruner when using multiplesbasis shapes


lducas:
The renormalization trick is not expected to be compatible with multiple basis (but this escape our naive test case). Please extend test case, investigate and repair.

PR #135, opened 2016-07-07
repaired renormalization bug

lducas:
renorm is determined by the first shape only


lducas:
This is supposed to repair issue #134

PR #139, opened 2016-07-15
convention on parameter ordering

malb:
typically output params are provided before input params, cf. GMP/MPFR APIs.


malb:
@lducas can you check if you're happy with this change?

PR #155, opened 2016-07-21
Pruner interface changes for fpylll

malb:
- highlevel `prune` function accepting many bases to average over
- change `vector<vector<double>> gso` to `vector<vector<double>*> gso` and `vector<MatGSO> gso` to `vector<MatGSO*> gso` to avoid copying
- instantiate templates

### DIFF
--- a/fplll/pruner.cpp
+++ b/fplll/pruner.cpp
@@ -35,7 +35,7 @@ template <class FT> void Pruner<FT>::set_tabulated_consts()
 
 template <class FT>
 template <class GSO_ZT, class GSO_FT>
-void Pruner<FT>::load_basis_shape(MatGSO<GSO_ZT, GSO_FT> &m, int start_row, int end_row, int reset_renorm)
+void Pruner<FT>::load_basis_shape(MatGSO<GSO_ZT, GSO_FT> &m, int start_row, int end_row, int /*reset_renorm*/)
 {
   if (!end_row)
   {
--- a/fplll/pruner.cpp
+++ b/fplll/pruner.cpp
@@ -90,10 +90,10 @@ template <class FT> void Pruner<FT>::load_basis_shape(const vector<double> &gso_
 }
 
 template <class FT> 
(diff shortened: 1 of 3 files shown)


malb:
@cr-marcstevens is my switch from `vector<MatGSO>` to `vector<MatGSO *> sound? It's actually quite inconvenient on the Cython side of things so if pointers could be avoided I'd be all too eager to avoid them.

cr-marcstevens:
Pointer arguments are horrible, but const vector<vector<double>>& or const vector<MatGSO>& may not be a bad argument type if you consider the use cases.
You can avoid copies of vector<double> or MatGSO by using move assignment, and vector<MatGSO> is something you likely have anyway if you're going to use this.
If you want to support other containers like array or c-arrays or lists you could have a templated function that takes a begin and end iterator.

malb:
I see your point for `vector<vector<double>>` but `vector<MatGSO>` would typically collect MatGSO objects of existing BKZ/LLL reductions. I'll revert the change for the `vector<vector<double>>` part, though.

cr-marcstevens:
Hi Martin,

If not for MatGSO then you're still likely to have some vector of objects from which you take the matgsos right?
Wouldn't it make sense to add a variant for that?

On July 21, 2016 6:37:24 PM GMT+02:00, Martin Albrecht <notifications@github.com> wrote:
>I see your point for `vector<vector<double>>` but `vector<MatGSO>`
>would typically collect MatGSO objects of existing BKZ/LLL reductions.
>I'll revert the change for the `vector<vector<double>>` part, though.
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/fplll/fplll/pull/155#issuecomment-234310784

-- sent from phone

malb:
Hi Marc, I had seen the wisdom of your ways and reverted my pointer-fication. For now, the main interface is Python and here we don't care too much about performance at the level of copy vs. move of a few objects.

Issue #160, opened 2016-07-24
option for not compiling libfplll


shi-bai:
The libtool seems to take a long time for compiling libfplll. In developing or debugging, each time making some changes and recompile, I have to wait for about ten secs on my laptop to wait for the compilation. It seems that libtool part can not be speeded-up by make -jN.  Thus I would prefer if there is an option in makefile.am to control whether we are compiling fplll as a library; or just compiling the binaries. The later could be useful in development stages. I could assign this to myself -- would like to see if everyone is happy with such feature (or dislike) :)

malb:
Mhh, are you sure it's actually libfplll which is killing performance, i.e. wouldn't you have the same problem also for the binary? If so, which step is slow? Is it linking? Maybe we can do something about it?

In any case, `--disable-libfplll` wouldn't hurt, so no objection :) I'm mainly trying to see if we can solve the problem instead of working around it.

cr-marcstevens:
I believe this problem is best solved via make targets, which already exist.
I.e., if you go into dir fplll and do 'make fplll' it should do the minimal (re-)compilation to build fplll.
Just 'make' builds the first target in Makefile, typically a 'all' target that compiles everything.

But I've you look at this latest line when building:

/bin/sh ../libtool  --tag=CXX   --mode=link g++-5.3.0 -std=c++11  -O3 -Wall   -o fplll main.o libfplll.la -lmpfr -lmpfr -lgmp -lgmp

Then it seems you actually need to compile the library to build the executables?

-- Marc

shi-bai:
On Sun, Jul 24, 2016 at 1:57 PM, cr-marcstevens <notifications@github.com>
wrote:

> I believe this problem is best solved via make targets, which already
> exist.
> I.e., if you go into dir fplll and do 'make fplll' it should do the
> minimal (re-)compilation to build fplll.
> Just 'make' builds the first target in Makefile, typically a 'all' target
> that compiles everything.
>
> But I've you look at this latest line when building:
>
> /bin/sh ../libtool --tag=CXX --mode=link g++-5.3.0 -std=c++11 -O3 -Wall -o
> fplll main.o libfplll.la -lmpfr -lmpfr -lgmp -lgmp
>
> Then it seems you actually need to compile the library to build the
> executables?
>
I think libfplll is pretty much used everywhere in compiling the
executables by looking into makefile.am. Thus a change of, e.g. nr/lll/bkz
will trigger a change of libfplll.



> -- Marc
>
[...]

PR #168, opened 2016-07-25
nothing depends on fplll.h

malb:
fplll.h includes everything so anything that includes it is recompiled
whenever anything changes, which can be slow.

See #160


shi-bai:
reviewed; also checked by linking lll/bkz from the old test.cpp.

PR #178, opened 2016-08-02
Cut releases from tags

malb:
This PR also enforces that `make distcheck` passes for future check-ins.


malb:
Merging myself because I seem to be the autoconf guy around here.

PR #183, opened 2016-08-14
Adding the NelderMead descent Method

lducas:
The pruner now offers 
- a more robust descent method NelderMead see https://en.wikipedia.org/wiki/Nelder%E2%80%93Mead_method, 
- as well as a hybrid, running Gradient first, and NelderMead then (should be faster while still robust).

Default method is the hybrid.


fplll/pruner.cpp
```diff
@@ -68,6 +33,32 @@ double svp_probability(const vector<double> &pr){
   return pru.svp_probability(pr);
 }
 
+
+
+// template <class FT> Pruner<FT>::Pruner()
```
malb: Maybe this can be removed?

lducas: Done.

PR #186, opened 2016-08-24
clean up soname handling and bump library version

malb:
Let's not merge this until @SnarkBoojum had a chance to tell us off or give us the thumbs up.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2,8 +2,6 @@
 # Process this file with autoconf to produce a configure script.
 
 AC_PREREQ(2.61)
-# don't forget to update library version in Makefile.am before a release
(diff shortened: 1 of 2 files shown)


SnarkBoojum:
It does bump the soname to 2, which is good.

I notice that you made so things have to be edited in both configure.ac and Makefile.am when a new version comes out ; that might not be very good. Here is how it is done in [eclib](https://github.com/JohnCremona/eclib/blob/master/configure.ac) and [pynac](https://github.com/pynac/pynac/blob/master/configure.ac) -- with Makefile.am modified to make use of what is defined in configure.ac.

malb:
Makes sense, fixed accordingly.

PR #188, opened 2016-08-25
Two typos and a leaking pkgconfig file

SnarkBoojum:
Those two commits do what is described in the title.

Thanks,

### DIFF
--- a/fplll.pc.in
+++ b/fplll.pc.in
@@ -6,5 +6,4 @@ includedir=@includedir@
 Name: @PACKAGE_NAME@
 Description: lattice algorithms with floating-poin computations
 Version: @PACKAGE_VERSION@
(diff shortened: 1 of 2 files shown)


malb:
Looks good, thanks. Let us know if you need a new point release for packaging (I don't want to proactively create too many of them)

SnarkBoojum:
I don't **require** a point release, but it would be much cleaner. Thanks!

malb:
Okay, maybe let's wait for a week or so to see if there are more issues that should be addressed for packaging etc.

SnarkBoojum:
I have pushed a 5.0.1 package to the experimental part of Debian, so when projects using fplll will have new versions supporting the new fplll, their packager will have an easier time.

If you publish new sub-releases, I'll see them and push them likewise into experimental.

Thanks!

PR #191, opened 2016-08-29
CVP improvements

gbonnoron:
Hi all,
The "resetting" feature of CVP has been implemented.
It solves the numerical problems that can occur in enumeration when the Gram-Schmidt coefficients are "unbalanced". The CVP should no longer be considered "experimental".

Now, the whole enumeration (Enumeration::enumerate()) may be recursively called, so the size of the EnumerationBase object matters. I changed some of its members that were std::array (static on the stack), into C-like arrays (dynamically allocated on the heap). The time cost for this change is roughly 1-1.5 %. It is manageable with the FPLLL_BIG_ENUM flag, enabled by default.


malb:
Cool, I'm sure @cr-marcstevens has something to say on the use of C arrays :smile:  I'd consider this a 5.1.x improvement, i.e. it'd warrant an increase of the version number by more than 0.0.1

cr-marcstevens:
Hi Guillaume,

Using dynamic C-arrays with all the new and delete calls makes the code more prone to errors and memory leaks
(besides a small performance penalty).
What is the exact problem you're trying to solve with this?

-- Marc

cr-marcstevens:
Is this about saving total memory usage, or about preventing stack overflow?

In the latter case, might it not be more convenient
to create the recursive Enumeration object etc in the new function reset_rec dynamically
instead of making all member variables of the Enumeration object dynamically allocated?
Then at most 1 Enumeration object might live on the stack,
but the new and delete of the Enumeration object would be within the same function.

If you use std::unique_ptr then you don't even need to worry about an explicit delete:
void test()
{
	std::unique_ptr<Foo> p1(new Foo); // create new object and store pointer in var p1
} // var p1 is destroyed which calls delete with the internal pointer.

cr-marcstevens:
Hi Guillaume,

This is indeed the idea:

   unique_ptr< Enumeration > enumobj( new Enumeration(_gso, new_evaluator));
   enumobj->enumerate(0, d, new_dist, 0, target, partial_sol, pruning_bounds, false, true);

The Enumeration object then doesn't live on the stack, but on the heap.
The new call to enumerate might cause a reset, but that reset will also create an Enumeration object on the heap.
So now all Enumeration objects live on the stack, except the first.

The only problem I now see is that that first Enumeration object might be quite large when you set max_enum_dim=1024,
then it will be several MBs large and might cause a stackoverflow by itself.

I still think it is better to make the entire object (EnumerationBase or Enumeration) dynamically allocated instead of all its member variables.
Even if it's okay now, the code is maintained by several people and bugs can easily creep in eventually.

Solutions:
- set max_enum_dir < 1024, for dim 200, setting it to 256 is fine.
- also create the first Enumeration object dynamically like the above code
- we rewrite Enumeration and EnumerationBase so that the large EnumerationBase is dynamically allocated.
  That means that EnumerationBase cannot be a base class anymore of Enumeration
  and that the calls to process_solution and process_subsolution have to be changed from a virtual call method to a callback method.
  I think this solution is actually better than what we currently have.

-- Marc

gbonnoron:
Hi Marc,

We can try all that. But I still need a very big enum size, I launch CVP in dim 600+ these days:) This is also why I may need to get rid of the std:array anyway. (they still live on the stack even if the object is on the heap, right?) What do you think?

Concerning the class hierarchy modification, what would that give us? Enumeration would contain a EnumerationBase object inside?

-- Guillaume

cr-marcstevens:
Hi Guillaume,

I've created a branch enumwrapper in fplll that ensures that all enumeration objects are allocated on the heap.
The new Enumeration class is a simple thin wrapper that contains a single pointer to an instance of the old Enumeration class.

Maybe you can try if that solves your problem.

Where the variable lives is essentially independent of the type of the variable.
std::array is a C++ wrapper around simple C-arrays like 'int x[100]',
but like those C-arrays and any other variable,
the variable can be either on the stack or on the heap:
- a global variable => heap
- a variable inside a function => stack
- dynamically allocated (new / delete) => heap
- a variable inside a struct/class => where the struct/class variable lives (see above 3 options)

-- Marc

gbonnoron:
Hi Marc,
Seems great :)
I merged your wrapper with my 'cvp' branch and committed it here.

cr-marcstevens:
Hi Guillaume,

I noticed you have code depending on k.
As you noticed that's an internal variable and not even used within the recursive enumeration.
So I don't think its a good idea to reference it for other rec/nonrec-independent functions.

For subsolutions I just pass either k or kk as argument 'offset'.
You do the same for reset_rec, but instead of also using that for non-recursive enum
you've made a copy without a function argument that directly refs k.
I think its better to use reset_rec in both cases and avoid code duplication.
The non-recursive code might be deprecated and removed sometime anyway.

-- Marc

On August 30, 2016 4:36:18 PM GMT+02:00, Guillaume Bonnoron <notifications@github.com> wrote:
>Hi Marc,
>Seems great :)
>I merged your wrapper with my 'cvp' branch and committed it here.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/fplll/fplll/pull/191#issuecomment-243460508

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

gbonnoron:
Hi Marc,
Indeed, much better this way :)

Speaking about code duplication: in the while loop in enumerate_recursive, there is a bunch of ~30 lines that are duplicated in the if statement on pardist[kk] != 0.0 Would it be better just to condition the updates of x, dx and ddx ?
I have also a suggestion on "x vs alpha" depending on dualenum. There are lots of code duplication due to that too. It could be better with a unique pointer set to x or alpha once and for all, don't you think?

--Guillaume

cr-marcstevens:
Hi Guillaume,

Good to hear!
Thanks for your efforts!

Just a last question:
Is there a reason why resetting is only supported for Float, and not for all possible types for FT?

BTW that code part is time critical, it is essentially executed for every node of the enumeration tree.
I explicitly duplicated it to increase performance.

-- Marc

On August 30, 2016 5:41:37 PM GMT+02:00, Guillaume Bonnoron <notifications@github.com> wrote:
>Hi Marc,
>Indeed, much better this way :)
>
>Speaking about code duplication: in the while loop in
>enumerate_recursive, there is a bunch of ~30 lines that are duplicated
>in the if statement on pardist[kk] != 0.0 Would it be better just to
>condition the updates of x, dx and ddx ?
>I have also a suggestion on "x vs alpha" depending on dualenum. There
>are lots of code duplication due to that too. It could be better with a
>unique pointer set to x or alpha once and for all, don't you think?
>
>--Guillaume
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/fplll/fplll/pull/191#issuecomment-243483253

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

cr-marcstevens:
BTW note that dualenum is a template argument so the 'if' is completely removed compiletime.
Using a pointer like you suggest would increase runtime code and take another CPU register.

On August 30, 2016 5:41:37 PM GMT+02:00, Guillaume Bonnoron <notifications@github.com> wrote:
>Hi Marc,
>Indeed, much better this way :)
>
>Speaking about code duplication: in the while loop in
>enumerate_recursive, there is a bunch of ~30 lines that are duplicated
>in the if statement on pardist[kk] != 0.0 Would it be better just to
>condition the updates of x, dx and ddx ?
>I have also a suggestion on "x vs alpha" depending on dualenum. There
>are lots of code duplication due to that too. It could be better with a
>unique pointer set to x or alpha once and for all, don't you think?
>
>--Guillaume
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/fplll/fplll/pull/191#issuecomment-243483253

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

gbonnoron:
That is right, all FT should get reset:)

malb:
Looks like we need to rebase. Other than that, is this ticket good to go?

gbonnoron:
For me, it is!

malb:
Okay, cool. Waiting for Marc's final :+1:  I can then attempt the merge, but feel free to beat me to it :smile:

cr-marcstevens:
Sure, but here are some items that may need further looking into:
- Is distinction for get_new_dist needed: can all possible FT also use the Float-specific code? Martin?

- use templated bool to enable/disable CVP-specific code for speed (like dual & findsubsols)
  instead of `!_max_indices.empty() && ...`

- better passing of arguments: e.g., the following always causes a copy for non-empty max_indices:
  `Enumeration(....., vector<int> max_indices=vector<int>())`

  recommended: `const vector<int>& max_indices = vector<int>()`

- more efficient test for reset, i.e., _max_indices never seems to be adjusted, so
the test `_max_indices[kk+1] == kk+1` can as well be a test `_max_indices_bool[kk+1] == true`. 
  or even better: a single integer for the first kk+1 for which `_max_indices[kk+1]==kk+1`
  e.g., `if (enablereset && kk = resetdepth) { ....`

We can also try to put it in a branch first for further refinement.

gbonnoron:
The previous commits implement all the comments:
- the get_new_dist can indeed be merged
- enable_reset is now handled as a template parameter
- handling the reset depth like suggested caught a bug I had introduced with earlier refactoring. It is now correctly used.

However this (now correct) "exact CVP" brings quite an overhead. In some cases, below the reset bounds, lots of nodes are visited compared to when doing the non-reset descent. I am thinking of a way to have both Exact and Heuristic easily callable. Test 1-3 should be for heuristic version and 4-5 for the exact variant. So more commits to come :)
Marc, what do you think about the modifications?

cr-marcstevens:
Very nice!
I couldn't make the changes in your repo.

With 'Heuristic' you mean without resetting?

cr-marcstevens:
:+1: :+1: :+1:

gbonnoron:
Here we go, I followed the same logic as in SVP, except for the default choice that I keep on fast variant. What do you think?
Martin, I will try to give you write access, otherwise what should be done?

gbonnoron:
It seems the make distcheck fails because of the environment in which it runs doesn't have the lattices... (e.g. tests/lattices/example_cvp*)

malb:
Hi, you need to add your example files to `EXTRA_DIST` in `Makefile.am` in `tests`, see https://github.com/fplll/fplll/blob/master/tests/Makefile.am#L8

gbonnoron:
;-)

malb:
Okay, everybody happy with this being merged as is?

shi-bai:
Good for me :)

Shi

malb:
Okay, merge time then.

PR #194, opened 2016-08-30
echo back chosen compiler for greater clarity

malb:
suggested by Marc via e-mail


malb:
@cr-marcstevens seems like we cannot disable the `gcc --version` thing. So I did the next best thing™ and also added `$CXX --version`.

PR #197, opened 2016-09-01
run self-dual BKZ in at least dd precision

### LINKED ISSUE #170: --march=native -O3
Compiling with `--march=native -O3` as per #169 leads to:

```
./test_bkz
terminate called after throwing an instance of 'std::runtime_error'
  what():  infinite loop in babai
[1]    8061 abort      ./test_bkz
```

on my machine.

### LINKED ISSUE #195: test_bkz in 5.0.1 fails on many architectures
Hi,

the Debian build bots found a problem with 5.0.1, see [full log for armhf](https://buildd.debian.org/status/fetch.php?pkg=fplll&arch=armhf&ver=5.0.1-1&stamp=1472595479) ; the error message is:

    terminate called after throwing an instance of 'std::runtime_error'
    what():  infinite loop in babai
    /bin/bash: line 5: 18226 Aborted                 ${dir}$tst
    FAIL: test_bkz

to see the list of failing architectures --all with the above error message, you can have a look [at this page](https://buildd.debian.org/status/package.php?p=fplll&suite=experimental).

I hope that helps,

Snark on #debian-science

malb:
fixes #170 and might possibly fixes #195

### DIFF
--- a/tests/test_bkz.cpp
+++ b/tests/test_bkz.cpp
@@ -237,13 +237,12 @@ int main(int /*argc*/, char** /*argv*/)
   int status = 0;
   status |= test_linear_dep();
   status |= test_filename<mpz_t>("lattices/dim55_in", 10, FT_DEFAULT, BKZ_DEFAULT | BKZ_AUTO_ABORT);
-  status |=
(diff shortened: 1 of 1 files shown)


shi-bai:
It works with -march=native -O3.

Please also refer to #170 and possibly #195

PR #201, opened 2016-09-03
GCC 4.8 Support

malb:
Turns out we can support GCC 4.8 as discovered by @SnarkBoojum  at https://trac.sagemath.org/ticket/21221


SnarkBoojum:
I don't think I had anything to do about it : be sure to send the flowers to the right person! From the trac records (pun intended), I'd say Jean-Pierre Flori is the one you wanted to thank.

malb:
Argh, sorry for the confusion.

PR #215, opened 2016-10-16
Add strict adherence to coding style in Travis

tlepoint:
I suggest to add a **coding convention check** to Travis: a build fails if the coding convention described in CONTRIBUTING.md is not used. The coding convention is enforced by running clang-format -i on all the .cpp/.h files: if one file is modified, then Travis exits with an error.

Instead of verifying all the modifications, focus on the first two commits:

- The build following the second commit [exited with errors](https://travis-ci.org/tlepoint/fplll/builds/168115342), as expected.

- The third commit corresponds to running `clang-format -i`. The build then [passed](https://travis-ci.org/tlepoint/fplll/builds/168115673).

If the behavior of exiting Travis with an error is undesirable, it could be run as an after-script and not impact the travis result (but it becomes less useful because it would still output an error in all builds for which the coding convention is not respected).


malb:
I'm a big fan of such checks/enforcement, but not everybody might be. I know it's risky to let this linger (bitrot) but we should wait for a bit before merging so that the others can state their preference.

Does the checker support some pragma/comment trigger to disable it for certain parts of the code? The generic rules are good in general but there are situations where manual formatting makes the code much more readable.

tlepoint:
Yes! http://llvm.org/releases/3.9.0/tools/clang/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code

malb:
Excellent.
- Can you add some documentation to `contributing.md` on those checks and manually disabling it?
- I'll ask on [fplll-devel]; bike-shedding is a dangerous issue here, but we're all adults :)

shi-bai:
I just tried the check-style branch. The changes are good to me. I had on my debian that the clang-format tool could be installed without installing clang, which is perhaps good for people who do not want to install clang.  Is this true in general?

malb:
Maybe `make check-style` (tab completion friendly) or `make style-check` would be more idiomatic?

tlepoint:
Done.

malb:
Looks good to me. Let's merge

PR #230, opened 2016-12-22
Stable lll in bkz

dstehle:
Enables babai (i.e., LLL size-reduction) not  to use the first basis vectors.
Changes enumeration post_processing in BKZ
    * if the solution vector contains a 1, use this to eliminate the linear dependency
    * if not, proceed as before

### DIFF
--- a/fplll/bkz.cpp
+++ b/fplll/bkz.cpp
@@ -126,7 +126,7 @@ bool BKZReduction<FT>::svp_postprocessing(int kappa, int block_size, const vecto
 {
   // Is it already in the basis ?
   int nz_vectors = 0, i_vector = -1;
-  for (int i = 0; i < block_size; i++)
+  for (int i = block_size-1; i >=0; i--)
   {
     if (!solution[i].is_zero())
     {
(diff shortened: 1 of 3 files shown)


shi-bai:
This was checked for dim 100 and 140, with blocksizes 2, 10, 20, 40, 60.

fplll/bkz.cpp
```diff
@@ -158,7 +178,7 @@ bool BKZReduction<FT>::svp_postprocessing(int kappa, int block_size, const vecto
     }
     m.row_op_end(d, d + 1);
     m.move_row(d, kappa);
-    if (!lll_obj.lll(kappa, kappa, kappa + block_size + 1))
+    if (!lll_obj.lll(0, kappa, kappa + 1, 0))
```
malb: Shouldn't this go to kappa + block_size + 1 to remove linear dependencies?

malb:
As far as I can tell this doesn't use the new truncated size reduction yet, but only changes post processing to run LLL less, right?

shi-bai:
oh, sorry, i don't know why " shi-bai merged commit dfa6b8a into master 2 minutes from now". what happened? did I click the merge button? .... ...

shi-bai:
> As far as I can tell this doesn't use the new truncated size reduction yet, but only changes post processing to run LLL less, right?
 Yes, Martin. The other one is not stable yet.

malb:
Looks like you merged by accident, we should fix that bug then :)

fplll/bkz.cpp
shi-bai: yes, sorry. let me up this.

shi-bai:
yes, I even mentioned that "let's leave it for a few days there for test"
.. don't now what had happened to me ...

PR #233, opened 2016-12-23
Pruner

lducas:
The gradient descent was going rogue due to inappropriate lower bounds on b[i]. Also, the descent now starts from greedy if reset=true (this way, it is also at least as good as greedy... sometime it wasn't !)


tests/test_pruner.cpp
```diff
@@ -229,11 +231,13 @@ template <class FT> int test_unpruned()
   pru.enumeration_radius = .85;
   double cost            = pru.single_enum_cost(pr);
   cerr << "Cost per enum " << cost << endl;
-
-  status       = (abs(1 - cost / 3.20e+10) > .02);
+  status |= (abs(1 - cost / 3.20e+10) > .02);
+  status |= std::isnan(cost);
+  cerr << std::isnan(cost) << endl;
```
malb: Do we want these kind of outputs here? They don't mean much to most people, I'd say.

tests/test_pruner.cpp
```diff
@@ -298,116 +303,234 @@ template <class FT> int test_auto_prune(size_t n)
             PRUNER_METRIC_PROBABILITY_OF_SHORTEST, true);
   status |= !(pruning.expectation <= 1.0);
   cerr << "Probability " << pruning.expectation << endl;
+  cost = 0.;
+  for (size_t i = 0; i < 2 * n; ++i)
+  {
+    cerr << pruning.detailed_cost[i] << "\t";
```
malb: Same printing question here.

tests/test_pruner.cpp
```diff
@@ -298,116 +303,234 @@ template <class FT> int test_auto_prune(size_t n)
...
+    cost += pruning.detailed_cost[i];
+  }
+  cerr << endl << "Total Cost " << cost << endl;
+
   status |= !(pruning.expectation > 0.0);
   status |= !(pruning.radius_factor >= 1.0);
   status |= !(pruning.coefficients[0] == 1.0);
 
+  if (status)
+    exit(1);
```
malb: We shouldn't exit(1) here but at least return 1. Why do the early abort at all? This prevents later tests from being run.

tests/test_pruner.cpp
lducas: Well, everything is silented out when check passes, right ?

tests/test_pruner.cpp
lducas: While, I want to know what exactly fails when it does... Can remove if bad practice.

tests/test_pruner.cpp
malb: If you run it through `make check` but not if you run it manually. I'd prefer outputs which the user can understand, i.e. which helps to resolve potential bugs. Or is that unreasonable to expect here?

tests/test_pruner.cpp
malb: My preference would be to print an error if a test failed and to continue. A bit more boilerplate code but it would give the maximum amount of feedback.

tests/test_pruner.cpp
lducas: Ok, removing the isnan output since the NaN is displayed anyway. May I leave all the data for which the value is preceded by what it is ?

malb: Sounds good to me. It's ultimately your decision in either case, I'm just providing (hopefully useful) feedback.

tests/test_pruner.cpp
lducas: Better, I'm printing a counter of errors after each of them ;)

PR #236, opened 2017-02-23
Add Gram-matrix object

kodebro:
**Code**
- Created gsob.cpp and gsob.h. These define the base GSO class (interface class): MatGSOb.
- Created ggso.cpp and ggso.h. These define a derived class MatGGSO from the MatGSOb. The class is based on the gram matrix.
- Mutated gso.cpp and gso.h, such that MatGSO is a derived class from MatGSOb.
- Mutated lll.cpp and lll.h, replacing every MatGSO by MatGSOb (and made some other small changes so that direct calls to the basis are avoided).

**Tests**
- Created test_gso.cpp, which tests the MatGSOb/MatGGSO/MatGSO object. It calculates the Gram-Schmidt version of a basis and a grammatrix which define the same lattice. After this, the testcode checks whether their r-matrices are equal. This can only be done for small matrices - in the case of large matrices, rounding errors will arise.
- Created test_lll_gram.cpp, which tests the lll.h and lll.cpp on the new object MatGGSO. It calculates both the LLL-reduced version of a basis and a grammatrix which define the same lattice:
LB = LLL(B)
LG = LLL(G)
Afterwards, LG is compared with LB^t* LB.

**Makefiles**
- Mutated fplll/Makefile.am, such that gsob/ggso can be built
- Mutated tests/Makefile.am,


malb:
Can we use more meaningful names following the rule that code is more often read than written? For example,
- `gsob.*` to  `gso_base.*`
- `ggso.*` to `gso_gram.*` (it's easy to overread the two consecutive g's)
- `MatGSOb` to `MatGSOBase` or `MatGSOInterface`
- `MatGGso` to `MatGramGSO` or `MatGSOGram`

malb:
Oh, we should probably also run some performance regression tests, say, some LLL and some BKZ.

kodebro:
I totally agree with the name suggestions. Are there standard performance tests present?
By the way - the `make check` runs correctly, but the `make distcheck` fails in Travis-CI. I am sorry to admit that I don't know what `make distcheck` means.

malb:
Hi, 

`make distcheck` checks if the tarball produced by `make dist` works. I'm fairly sure your problem comes from not adding your new test matrix to `EXTRA_DIST=` in `tests/Makefile.am`. Thus, it is not included in the tarball and `make distcheck` fails. You can run `make distcheck` locally, to see if that fixes it.

We, unfortunately, don't have any standard performance tests. But we should! Maybe some LLL and some BKZ-60 on some `double` and some `mpfr` matrices?

kodebro:
(1) Including the new testmatrix indeed removes the build interruption. However, it turns out that `make check` gives 9 PASSes, but `make distcheck` doesn't (my two new test files give a FAIL). I will look at this.
(2) I will make an issue for constructing standard performance tests; I think also some numerical stability tests are wanted in the future.

kodebro:
It seems that, during `make distcheck`, the test lattices are not in the same folder as the built test files. When doing `make distcheck`, one gets a temporary folder `fplll-5.1.0/_built/tests/` where the lattices are _not_ included.

After doing `make distcheck` on the _current master branch_ of fplll, and interrupting directly after the 7 PASSes, I found that `test_lll.log` contains 7 times:
> is_lll_reduced reports success when it should not
which strongly suggest that the input matrices are empty.

Is it supposed to be this way?

malb:
This was indeed a bug. I've just fixed it and fixed up the `test_lll` tests, in dc3d2eb you see how to fix up your failing tests when called from `make distcheck`

lducas:
Seems like the travis error is merely a clang vs clang-3.9 issue. Should this be ignored ?

malb:
I agree, just merged a config to ignore it.

malb:
Hi, as far as I can tell, these would be the next steps:
- you should rebase on the current master, this should shut up travis
- @shi-bai will give his feedback as well
- We should have some performance numbers to check. Doesn't have to be too much just some LLL on master and on this branch to check there isn't something really strange going on.

Does that sound about right?

shi-bai:
Hi,

On Sat, Mar 11, 2017 at 6:55 AM, Martin Albrecht <notifications@github.com>
wrote:

> Hi, as far as I can tell, these would be the next steps:
>
>    - you should rebase on the current master, this should shut up travis
>    - @shi-bai <https://github.com/shi-bai> will give his feedback as well
>    - We should have some performance numbers to check. Doesn't have to be
>    too much just some LLL on master and on this branch to check there isn't
>    something really strange going on.
>
> It looks good to me. I was also doing some performance test. So far I have

n = dimension, average timing of current fplll's lll, average timing of
purge request's lll (over 50 trials each dimension)
n=100  80.5,   80.9
n=200  2733.8  2909.2
n=250  8060.8  8612.8
n=300  20143.7 20468.3

I am looking further into n=250 to see what makes the difference.

Cheers,
Shi

shi-bai:
Hi,
I did some more comparison for n=250 (on a different platform to the previous comparison -- but the same for the two tests in this comparison) and have the following timings

n=250 7675.62 7851.52

Koen, do you know what could be the possible reason for the difference?

Attached one example(s) on the timing comparison (lattice generated by the svp challenge generator with n=250 and bit = 100 * n).

Also the runs with current fplll and purge request's lll ended up with the same reduced lattices (thus correctness is good).

Just a small thing on the naming consistency, perhaps could use gso_interface.c instead of gso_base.c; or use MatGSOBase in gso_base.c?

[n250_lll.txt](https://github.com/fplll/fplll/files/837212/n250_lll.txt)
[n250_lll_gram.txt](https://github.com/fplll/fplll/files/837213/n250_lll_gram.txt)

kodebro:
Hey,
Thanks for reviewing!

Those timings doesn't look to great. I will try to find the cause.
- I will rename gso_base.* to gso_interface.*
- I will rebase on the current master.

kodebro:
This Wednesday I will analyse my code. 
@shi-bai , as I understand from your textfiles, the difference in efficiency also occurs when `enable_int_gram` is turned off. That shocks me a little, because I thought I didn't mutate (logically) the code where `enable_int_gram` is turned off. I will thoroughly look at this, Wednesday.

kodebro:
Hey all,
I took a look at it and I think I must conclude that I still think that I didn't change the code of gso.* logically, only structurally.

I had a short chat with @cr-marcstevens about this, and he suggested that the (lots of) virtual functions in the abstract class `MatGSOInterface` might be the cause of this significant slow-down. Might using templates be a solution?

lducas:
`` the difference in efficiency also occurs when enable_int_gram is turned
off. ''

This could also suggest lack of determinism due to Floating Point
Arithmetic ? We are speaking of running wrapper here ? If so there could be
timing differences induced by a run switching to highter precision earlier
than the other.

Best
-- Leo

kodebro:
From the files of @shi-bai, one can draw a few (careful) conclusions:
(1) The switching to higher precision occurs at the same time for the two versions. At precision 53 they both report failure at the 169th vector, and at precision 64 they both report failure at the 217th vector. With precision 106, they both succeed. This suggests that timing differences induced by switching earlier (at least in this particular case of n = 250) is ruled out. This is consistent with my observation that the code is not changed logically.
(2) The Gram-version is consistently slower (at nearly every discover-vector) than the original (master) version. Even for discovering the latest rows, the Gram-version takes roughly 1.2 times as much time as the original version. See below for the relative timings, computed by dividing the timing of a discover-row of the Gramversion by the timing of a discover-row of the masterversion.

timingsfplllrelative

lducas:
Just a sanity check: we are comparing old code with new code, but in both case we use the basis-LLL and not the gram-LLL, right ?

kodebro:
What I understood from it: Yes.

shi-bai:
> Just a sanity check: we are comparing old code with new code, but in both case we use the basis-LLL and not the gram-LLL, right ?
Yes, there is no gram-LLL involved. 

It looks to me the functionality for the two versions are the same. However the timing has been increased slightly -- might have something to do with the virtual functions. 

I have used the following flag for both comparison
DEFAULT_CXXFLAGS="-O3 -march=native"

Could someone check if you have a similar timings on your desktop/laptop?

kodebro:
Hey all,

I am repeating the experiments of @shi-bai right now, applying it to three versions: the master branch, this pull-request branch and a 'improved' pull-request branch. These improvements include marking a few virtual functions inside the `gso.h` file as `final`, in order to diminish the virtual-function overhead.

In smaller experiments I did (averaging on twenty dimension 200 matrices), this last version seemed to use 0.6% more time than the master branch. I will soon inform you how the timing differences appear to be related to the dimension.

kodebro:
Hey all,

Me and @shi-bai both tested various instances of this pull request. I will here state my results. I will state the (observed) average and standard deviation of the master-timings. Also I will state the (observed) average and standard deviation of the pull-request, and also the average of the differences (of the timings of master vs. pull request). Every dimension has 10 samples, except dimension 300, which has 4. Everything in seconds.

Dimension 100:
- Master :   mu = 19.914, sigma = 1.105,  sigma/mu = 5.6%
- PullReq: mu = 19.433, sigma = 1.262
- Difference = -0.480  ( - 2.41%)

Dimension 200:
- Master :   mu = 874.872, sigma = 10.784,  sigma/mu = 5.6%
- PullReq: mu = 884.720, sigma = 26.353
- Difference = 9.847  ( +1.13%)

Dimension 250:
- Master  : mu = 12170, sigma = 147,  sigma/mu = 1.21%
- PullReq: mu = 12208, sigma = 351
- Difference = 38  ( +0.32%)

Dimension 300 (only 4 samples):
- Master :  mu = 43195, sigma = 5049,  sigma/mu = 11.7%
- PullReq: mu = 44080, sigma = 4256
- Difference = 884  ( +2.0%)

shi-bai:
Koen thanks for more experiments. Shall we merge the request?

kodebro:
If you all agree, I think that is a good idea :)

lducas:
+1

malb:
+1

lducas:
I'll go ahead and just click the button then.

Issue #245, opened 2017-05-23
Tests on 32-bit


malb:
Sage [reports](https://trac.sagemath.org/ticket/23057#comment:1) that fplll doesn't pass self tests on 32-bit systems. In particular, `test_pruner` is failing.

lducas:
I will make the comparison a bit looser. But I'm not sure how to test things in a 32bit env. I don't have one, maybe I can set a virtual machine or something ?

malb:
Fixed.

PR #246, opened 2017-05-24
Repairing test_pruner for #245

lducas:
Inequalities in test_pruner now tolerates small numerical errors.
The test still pass on my 64 bits machine. Not tested yet on a 32 bit env.


malb:
Looks like it worked: https://trac.sagemath.org/ticket/23057

PR #251, opened 2017-06-06
Sieve api 248

shi-bai:
Does anyone know how to fix the error "codecov/patch — No report found to compare against"?


malb:
Just ignore it, I think it simply doesn't have a report for master yet. Looking good.

Issue #252, opened 2017-06-06
Avoid `exit(1)` calls


malb:
In https://github.com/fplll/fplll/blob/master/fplll/sieve/sieve_gauss.cpp#L133 `exit(1)` is called. This should be replaced by raising exception (or calling `abort()` the very least) so that higher level code can recover if it chooses to.

shi-bai:
I closed this. Please refer to another ticket #254

PR #256, opened 2017-06-27
Remove exit

shi-bai:
#252


shi-bai:
**Note: I removed the automake in brew install in travis since otherwise it returns me the following error:
error: automake 1.15 is already installed**

PR #268, opened 2017-07-12
Allow BOUNDED_LLL in Slide reduction and remove LLL from postprocessing

blowfish880:
- remove all lll calls from post processing (this results only in a very slight speed up, but the usefulness of this is rather conceptual, i.e. we prefer to have all the lll calls in the preprocessing)
- ensures lll is called in slide reduction even if BKZ_BOUNDED_LLL is on
- allows row operations through the gso even for pairs (i,j) s.t. i > j
- fixes a bug in the test_gso.cpp and corresponding issue in gso_gram.cpp (see point above)


fplll/bkz.cpp
```diff
@@ -447,13 +456,24 @@ bool BKZReduction<FT>::slide_tour(const int loop, const BKZParam &par, int min_r
...
   {
     clean = true;
-    // SVP reduction takes care of the LLL reduction as long as BKZ_BOUNDED_LLL is off
     for (int i = 0; i < p; ++i)
     {
       int kappa      = min_row + i * par.block_size;
       int block_size = min(max_row - kappa, par.block_size);
       clean &= svp_reduction(kappa, block_size, par);
     }
+    // SVP reduction takes care of the LLL reduction as long as BKZ_BOUNDED_LLL is off
```
malb: "as long as" → "if" (?)

fplll/bkz.h
```diff
@@ -61,9 +61,9 @@ template <class FT> class BKZReduction
 
   bool svp_preprocessing(int kappa, int block_size, const BKZParam &param);
 
-  bool svp_postprocessing(int kappa, int block_size, const vector<FT> &solution);
+  bool svp_postprocessing(int kappa, int block_size, const vector<FT> &solution, bool dual = false);
 
-  bool dsvp_postprocessing(int kappa, int block_size, const vector<FT> &solution);
+  bool svp_post_general(int kappa, int block_size, const vector<FT> &solution, bool dual = false);
```
malb: I'd call this `svp_postprocessing_generic`

PR #281, opened 2017-07-13
BKZ with Z_NR<long>

blowfish880:
bkz reduction runs with long integers instead of mpz, if the numbers in the basis are small enough. the bkz reduction is templated with an integer type, the switch happens in the bkz_reduction_f function, so the user interface (bkz_reduction(...)) did not change.

### DIFF
--- a/fplll/bkz.cpp
+++ b/fplll/bkz.cpp
@@ -26,8 +26,8 @@
 
 FPLLL_BEGIN_NAMESPACE
 
-template <class FT>
-BKZReduction<FT>::BKZReduction(MatGSO<Integer, FT> &m, LLLReduction<Integer, FT> &lll_obj,
+template <class ZT, class FT>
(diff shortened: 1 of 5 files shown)


fplll/bkz.cpp
```diff
@@ -748,11 +753,33 @@ int bkz_reduction_f(IntMatrix &b, const BKZParam &param, int sel_ft, double lll_
...
   if (sel_ft == FT_DOUBLE || sel_ft == FT_LONG_DOUBLE)
     gso_flags |= GSO_ROW_EXPO;
-  MatGSO<Integer, FT> m_gso(b, u, u_inv, gso_flags);
-  LLLReduction<Integer, FT> lll_obj(m_gso, lll_delta, LLL_DEF_ETA, LLL_DEFAULT);
-  BKZReduction<FT> bkz_obj(m_gso, lll_obj, param);
-  bkz_obj.bkz();
-  return bkz_obj.status;
+
+  ZZ_mat<long> bl;
+  if (convert<IntegerT, long>(b, bl, 10))
```
malb: Use `mpz_t` to be explicit here instead of `IntegerT`?

PR #288, opened 2017-07-14
Nr docs

drazioti:
Documentation for nr_rand

### DIFF
--- a/fplll/nr/nr_rand.inl
+++ b/fplll/nr/nr_rand.inl
@@ -11,7 +11,17 @@ FPLLL_BEGIN_NAMESPACE
 
 class RandGen {
 public:
-  static void init() {
+ /**    
+	@brief Create a random generator (using gmp):
+
+	@param init             Initialize State for a default algorithm (Mersenne Twister) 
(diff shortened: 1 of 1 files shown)


fplll/nr/nr_rand.inl
```diff
@@ -11,7 +11,17 @@ FPLLL_BEGIN_NAMESPACE
 
 class RandGen {
 public:
-  static void init() {
+ /**    
+	@brief Create a random generator (using gmp):
```
malb: These are not `params` but methods, each should get its own docstring

fplll/nr/nr_rand.inl
```diff
@@ -45,6 +55,13 @@ private:
 
 class RandGenInt {
 public:
+ /**    
+	@brief Create a Random Generator
+
+        @param init  Initialize a random number generator. It does not return somethinng     
+        @param get   Returns a pseudo-random integral number in the range between 0 and RAND_MAX (RAND_MAX is at least  32767)
+        @get_bit   Returns a random bit
```
malb: Same as above

PR #290, opened 2017-07-14
Class scheme

Vaggelis1994:
Added the class scheme image between the correlation of the library classes. Not sure if it should be submitted in the root project folder.


malb:
Let's add it to the wiki instead.

Vaggelis1994:
should the link be a reference to the project folder though?

malb:
I'm quite strongly opposed to adding this to the source repository. Information like that gets outdated quickly. I suggest to close this PR and to upload the class scheme to the Wiki instead.

malb:
cf. https://fplll.github.io/fplll/inherits.html and https://fplll.github.io/fplll/gso_8h.html

shi-bai:
As we have the doxygen online which could be updated continuously in the future, I agree it's better to have class_scheme.png in the wiki instead. I closed this pull request.

Vaggelis1994:
I agree with your points. Indeed, this information is updated regularly. Nevertheless, the wiki page needs a URL link to attach an image, and it is consequently not possible for someone to upload it directly over there. This is why I proposed to put the scheme in the project folder and create a link for wiki from there.

Do you have any other recommendations on that?

malb:
Actually, the Wiki is a Git repository itself, you can clone it by calling `git clone git@github.com:fplll/fplll.wiki.git` You could then add a new entry and include the file there. I've just made you a contributor, so you have push rights to the Wiki

PR #312, opened 2017-08-10
Pruner: Throwing errors when encountering NaN

lducas:
- throwing errors on NaN
- throwing error in infinite loop in descent_step
- additional printouts on flag PRUNER_VERBOSE

### DIFF
--- a/fplll/pruner.cpp
+++ b/fplll/pruner.cpp
@@ -16,145 +16,155 @@
 #include "pruner.h"
 #include "ballvol.const"
 #include "factorial.const"
-#include "gso.h"
+#include "fplll/nr/nr.h"
 
 FPLLL_BEGIN_NAMESPACE
 
-template <class FT> FT svp_probability(const Pruning &pruning)
+// PRIVATE METHODS
+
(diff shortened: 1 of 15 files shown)


fplll/pruner.cpp
```diff
@@ -289,6 +298,11 @@ inline FT Pruner<FT>::single_enum_cost(/*i*/ const evec &b, vector<double> *deta
     total += tmp;
     normalized_radius_pow *= normalized_radius;
   }
+  if (!total.is_finite())
+  {
+    throw std::runtime_error("NaN or inf in single_enum_cost");
```
malb: Some of these might be more appropriate http://en.cppreference.com/w/cpp/error/exception

lducas: hum. Would you consider this an overflow ?

malb: I was thinking that `is_finite` checks if the value is within representable range.

lducas:
K

Le 10 août 2017 11:56, "Martin Albrecht" <notifications@github.com> a
écrit :

Issue #340, opened 2017-12-29
randomness is machine dependent


malb:
In commit db534dcb70c0ed5f5b9bd018f69f60dae78cebd5 we introduced the following code:

```C++
  if (bits > 32){ 
    unsigned long int tmp = mpz_get_ui(data);  
    gmp_randseed_ui(RandGen::gmpState, tmp*tmp);
  }
```

with commit message "Attempt to break the eventual linearity of the GMP PRNG." A side-effect of this code is that random matrices now depend on whether the architecture is 32-bit or 64-bit (because we're going via a long).

1. Do we really need this workaround? What goes wrong if we remove it?

2. If we need it, can we do `tmp & ~(1<<31)` or so to unify the two families?

dstehle:
The reason for that tweak is that gnump's pseudo-random numbers are derived from a linear congruential generator. This could be bad for testing the behaviour of lattice reduction algorithms, as these algorithms are typically those that are used to break such pseudo-random number generators. 
It's not a cryptographically clean patch, but at least, by squaring tmp, every time we call 'rand' to get an integer, linearity with the previous call is broken. 

I don't know why "if (bits > 32)" is here. This potential linear dependancy occurs also when we ask for small random integers, not just long ones. I would remove this restriction.  

Why "unsigned long int"? It's just the type that  gmp_randseed_ui is waiting for. 
Your solution "tmp & ~(1<<31)" does the unifying job, I think. 
More precisely, adding a line tmp = tmp & ~(1<<31) just after the creation of tmp.

malb:
using this strategy I'm now seeing:

```shell
 ./test_lll_gram 
The gram GSO-object is not LLL-reduced after calling LLL
The gram GSO-object is not LLL-reduced after calling LLL
```

CC: @kodebro

kodebro:
This last issue should be solved by [https://github.com/fplll/fplll/pull/342](url).

Issue #328, opened 2017-11-12
Gram GSO LLL doesn't work with gram matrices with big coefficients (was: doesn't work with non-representable)


kodebro:
It seems that GramGSO LLL doesn't work on gram matrices that are *not* of the form G = B*B^T. I tried to put in the following matrix:
`[[3559941422478111945438014 2000384006369912794808478 2229599748450132568394343 749522257835011031240761]
[2000384006369912794808478 1124045510320513281506436 1252845243251596873063781 421167698862786742516112][2229599748450132568394343 1252845243251596873063781 1396403605660581339138906 469427566132091561402657][749522257835011031240761  421167698862786742516112  469427566132091561402657 157806982846091154484154]]`

Reduction of fplll:
`[ 154    0     0                                               0 ]
[  37 3816     0                                               0 ]
[  17 1711  9146                                               0 ]
[ -60 1614 -3036 -3672475935620874863122493963413516150045490214 ]`
Real reduction:
`[ 4 -1 -1  1]
[-1  4 -1  1]
[-1 -1  4  1]
[ 1  1  1  4]` 

Something is wrong here. I will make a branch in my cloned repository with a mutated testfile including this example soon.

malb:
FWIW Here's the same example reproduced in fpylll

```python
In [9]: A = IntegerMatrix.from_matrix([[3559941422478111945438014, 2000384006369912794808478, 2229599748450132568394343, 749522257835011031240761], [2000384006369912794808478, 1124045510320513281506436, 1252845243251596873063781, 421167698862786742516112], [2229599748450132568394343, 1252845243251596873063781, 1396403605660581339138906, 469427566132091561402657], [749522257835011031240761, 421167698862786742516112, 469427566132091561402657, 157806982846091154484154]])

In [11]: M = GSO.Mat(A, gram=True)
In [12]: LLL.Reduction(M)()
In [13]: print A
[                 58858964 2000384006369912794808478                  2229599748450132568394343 749522257835011031240761 ]
[                 11273758                2394353606                  1252845243251596873063781 421167698862786742516112 ]
[                 -7765351                -350937627 -84743144112504095273476788176669742654916 469427566132091561402657 ]
[ 749522257835011031240761  421167698862786742516112                   469427566132091561402657 157806982846091154484154 ]

```

kodebro:
Shouldn't there be an `update_gso()` somewhere inbetween?
When I tried this, I got
```
sage: M = GSO.Mat(A, gram = True)
[...]

Issue #262, opened 2017-07-10
test_lll_gram on rpi2


lgremy:
I compiled fplll, with gcc version 6.3.0 20170516 (Raspbian 6.3.0-18+rpi1), and obtained the following fail by running ```make check``` on my Raspberry Pi 2, with an ARMv7 Processor rev 5 (v7l) processor.
```
=======================================
   fplll 5.1.0: tests/test-suite.log
=======================================

# TOTAL: 9
# PASS:  8
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: test_lll_gram
===================

The gram GSO-object is not LLL-reduced after calling LLL
FAIL test_lll_gram (exit status: 255)
```

malb:
Can you run `./test_lll_gram` to see where it bombs out and how?

kodebro:
I had a quick talk with Laurent and Léo, and apparently the following lines are the troublemakers:
`status |= test_int_rel<mpz_t, double>(200, 100);`
` status |= test_int_rel<mpz_t, mpfr_t>(200, 100);`
It seems that the dimension (first argument) is way to high for the precision. As Léo pointed out to me, those dimensions are carefully chosen, depending on the precision. 

In my local repository, I deleted all `test_int_rel` with dimension 200. I think deleting those tests should do the trick.

I don't know (yet) an explanation why the test does work for MatGSO-objects but doesn't for MatGramGSO-objects.

lgremy:
In the attached file, I list matrices that failed, and a one that does not failed, because the matrices are randomly generated and I prefer to explicitly give them.
[test_lll_gram.txt](https://github.com/fplll/fplll/files/1135909/test_lll_gram.txt)

kodebro:
I investigated somewhat more, and found the bug. The main problem is that the full Gram matrix was not copied. I will fix this - the good news is that the error is (probably) not in the code of GramGSO.

kodebro:
I will very soon make a pull-request for this!

gilvillard:
[...]

PR #342, opened 2018-01-16
Remove 200-dimensional tests in test_gram_lll.cpp

kodebro:
Removed the 200-dimensional tests and cleaned up the code in test_gram_lll.cpp.

I found that taking a too big bit size without using Wrapper, causes overflow in Babai-rounding.


kodebro:
This should (partially) resolve
[https://github.com/fplll/fplll/issues/340](url)
[https://github.com/fplll/fplll/issues/262](url)
[https://github.com/fplll/fplll/issues/328](url)

PR #344, opened 2018-01-19
Basic HNF functionalities

arnsip:
-added functions to the core executable code to provide hnf usage for bash/shell
-added a way to create rectangular matrices for latticegen (was needed for my tests)
-added basic xgcd hnf algorithm, function to check if a vector is inside a lattice given a HNF, if a basis corresponds to a HNF, if a basis is a HNF at all (test_hnf.cpp provided and added to make check)

This is not the final version, don't use hnf on large matrices or large determinants (it's slow).

This is provided for future stability with fork/pulls from anyone (including me) and to give basic functionality at least.


arnsip:
Hum... I tried to run make check-style a lot successively without changing my code but it seems errors can get worse. Not sure what to do to make travis happy.

lgremy:
My two cents: version of clang-format. Can you try with `make check-style CLANGFORMAT=clang-format-3.9`?

fplll/bkz.cpp
```diff
@@ -347,9 +347,11 @@ bool BKZReduction<ZT, FT>::svp_reduction(int kappa, int block_size, const BKZPar
     throw std::runtime_error(RED_STATUS_STR[lll_obj.status]);
   }
 
-  // in order to check if we made progress, we compare the new shortest vector to the
+  // in order to check if we made progress, we compare the new shortest vector
```
malb: Why are these whitespace changes introduced here? I think these are a regression

fplll/bkz.h
```diff
@@ -110,8 +110,10 @@ template <class ZT, class FT> class BKZReduction
    * @param block_size
    *    size of the block
    * @param param
-   *    parameter object for the current block size (the parameter object for the recursive
-   *    calls will be created in this function using the information from this object)
+   *    parameter object for the current block size (the parameter object for
```
malb: Same whitespace issue as bkz.cpp

malb:
Hi, the patch touches many files I think it shouldn't and makes whitespace changes which I'd consider regressions.

fplll/defs.h
```diff
@@ -167,6 +168,7 @@ const char *const RED_STATUS_STR[RED_STATUS_MAX] = {"success",
                                                     "error in SVP solver",
                                                     "error in BKZ",
                                                     "time limit exceeded in BKZ",
+                                                    "error in HNF",
```
malb: The order seems wrong here.

fplll/hnf.h
```diff
@@ -0,0 +1,91 @@
...
+ * @brief Tests a couple matrix vector
+ *    returns 1 if B is not a HNF or -1 if v is not in L(B), else 0
+ *
+ * @param B
+ *    HNF basis of the lattice to be tested
+ * @param v
+ *    vector for the membership test
+ */
+
+int in_hnf(const ZZ_mat<mpz_t> &B, const vector<Z_NR<mpz_t>> &v);
```
malb: I find `in_hnf` misleading, why not ask `in_lattice`?

fplll/latticegen.cpp
```diff
@@ -41,7 +42,9 @@ void print_version()
   cout << "latticegen (fplll) " << VERSION << endl
        << "Copyright 2005-2011 Damien Stehle, David Cade, Xavier Pujol." << endl
        << "fplll is free software. You can redistribute it and/or modify" << endl
-       << "it under the terms of the GNU Lesser General Public License as published by" << endl
+       << "it under the terms of the GNU Lesser General Public License as "
```
malb: This looks a bit odd

fplll/main.cpp
```diff
@@ -1,6 +1,7 @@
 /* Copyright (C) 2005-2008 Damien Stehle.
    Copyright (C) 2007 David Cade.
    Copyright (C) 2008-2011 Xavier Pujol.
+   Copyright (C) 2018 Arnaud Sipasseuth (just the hnf part)
```
malb: I'd just write "(HNF)"

fplll/main.cpp
```diff
@@ -576,28 +625,35 @@ void read_options(int argc, char **argv, Options &o)
...
       exit(0);
     }
     else if (strcmp(argv[ac], "--version") == 0)
     {
       cout << "fplll " << VERSION << endl
            << "Copyright 2005-2012 Damien Stehle, David Cade, Xavier Pujol." << endl
            << "fplll is free software. You can redistribute it and/or modify" << endl
-           << "it under the terms of the GNU Lesser General Public License as published by" << endl
-           << "the Free Software Foundation, either version 2.1 of the License, or" << endl
+           << "it under the terms of the GNU Lesser General Public License as "
```
malb: This looks odd

fplll/main.cpp
```diff
@@ -576,28 +625,35 @@ void read_options(int argc, char **argv, Options &o)
...
+              "type)\n"
            << "  -bkzboundedlll\n"
            << "        Restricts the LLL call\n"
            << "  -bkzdumpgso <file_name>\n"
            << "        Dumps the log of the Gram-Schmidt vectors in specified file\n"
            << "  -of [b|c|s|t|u|v|bk|uk|vk]\n"
            << "        Output formats.\n"
 
-           << "Please refer to https://github.com/fplll/fplll/README.md for more information.\n";
+           << "Please refer to https://github.com/fplll/fplll/README.md for "
```
malb: This looks odd.

fplll/nr/matrix.h
```diff
@@ -266,19 +266,24 @@ template <class ZT> class ZZ_mat : public Matrix<Z_NR<ZT>>
   void gen_simdioph(int bits, int bits2);
   /** Generate a random matrix of uniform distribution. */
   void gen_uniform(int bits);
+  /** Generate a random matrix of uniform distribution (non-square flexible). */
```
malb: Why not permit non-square for `gen_uniform`? I don't think we need two functions for this.

fplll/nr/matrix.h
```diff
@@ -266,19 +266,24 @@ template <class ZT> class ZZ_mat : public Matrix<Z_NR<ZT>>
...
+  void gen_uniform2(int bits);
 
-  /** Construct a matrix `[[I,H],[0,qI]]` where `H` is constructed from rotations of a vector ``h``.
+  /** Construct a matrix `[[I,H],[0,qI]]` where `H` is constructed from rotations of a vector
+    ``h``.
```
malb: I'm not too keen on these line breaks, but I won't make them a show stopper

fplll/nr/nr_Z.inl
```diff
@@ -208,22 +208,39 @@ public:
...
   inline void randm_si(const Z_NR<Z>& max);
 
-  /**	 
-   * Generates smallest prime number above nbr 
+  /**
+   * Generates smallest prime number above nbr
   */
   inline void nextprime(const Z_NR<Z>& nbr);
 
+  /**
```
malb: I don't like this strategy: it leads to "flexible" templated code which isn't flexible at all. LinBox was guilty of that for a long time: you'd have to guess/know which templates can actually be instantiated. I'd rather have these for all data types unless there is a very good reason.

tests/Makefile.am
```diff
@@ -29,15 +29,15 @@ EXTRA_DIST=test_utils.h \
 	lattices/example_cvp_out2 \
 	lattices/example_cvp_out3 \
 	lattices/example_cvp_out4 \
-	lattices/example_cvp_out5 
+	lattices/example_cvp_out5
```
malb: Does your editor delete trailing whitespaces unconditionally?

tests/test_bkz.cpp
```diff
@@ -171,7 +171,8 @@ int test_bkz_param_pruning(ZZ_mat<ZT> &A, const int block_size, int flags = BKZ_
 }
 
 /**
-   @brief Test BKZ_DUMP_GSO for a matrix d × (d+1) integer relations matrix with bit size b (copied
+   @brief Test BKZ_DUMP_GSO for a matrix d × (d+1) integer relations matrix with
+   bit size b (copied
```
malb: Line breaks

fplll/hnf.cpp
```diff
@@ -0,0 +1,346 @@
...
+    cerr << "in_hnf error : matrix-vector sizes do not match\n";
+    return -1;
+  }
+  /* matrix indexes (k = pivot)*/
+  int i, j, j2, k;
+  /* the limit "l" depends of matrix dimensions */
+  int l = (c - r) * (c > r);
+
+  /* main loop, decreases from last column to the limit */
+  for (j = c - 1, k = r - 1; j >= l; j--, k--)
```
malb: Isn't this `is_hnf_reduced()`? Why copy'n'paste?

fplll/hnf.cpp
```diff
@@ -0,0 +1,346 @@
...
+}
+
+int in_hnf(ZZ_mat<mpz_t> &B, const ZZ_mat<mpz_t> &A)
+{
+
+  ZZ_mat<mpz_t> tmp = A;
+  Z_NR<mpz_t> q;
+
+  /* matrix bounds */
+  int r = B.get_rows(), c = B.get_cols();
```
malb: This looks like copy'n'pasted code again or am I missing something?

fplll/hnf.cpp
```diff
@@ -0,0 +1,346 @@
+/* Copyright (C) 2018 Arnaud Sipasseuth. (inspiration from FLINT (2017 version))
```
malb: The default API of fplll is "make an HNF object and then call that". I appreciate this isn't quite needed here but it makes sense to follow the API convention anyway (?)

fplll/defs.h
```diff
@@ -179,6 +181,15 @@ enum LLLMethod
...
 const char *const LLL_METHOD_STR[6] = {"wrapper", "proved", "heuristic", "fast"};
 
+enum HNFMethod
+{
+  HM_CLASSIC     = 0,
+  HM_XGCD        = 1,
+  HM_PERNETSTEIN = 2
+};
+
+const char *const HNF_METHOD_STR[6] = {"classic", "xgcd", "pernet-stein"};
```
malb: These aren't actually all implemented, right? We shouldn't advertise something which is not implemented yet. Also the array is too big.

tests/test_pruner.cpp
```diff
@@ -284,7 +286,8 @@ template <class FT> int test_unpruned()
   {
     v3[i] *= 20;
   }
-  // 1 basis out of 3 is so large that it essentially induces no cost: new cost should be 2/3 of
+  // 1 basis out of 3 is so large that it essentially induces no cost: new cost
```
malb: More whitespace changes.

tests/test_lll_gram.cpp
```diff
@@ -185,7 +185,8 @@ template <class ZT, class FT> int test_filename(const char *input_filename)
 }
 
 /**
-   @brief Construct d × (d+1) integer relations matrix with bit size b and test LLL.
+   @brief Construct d × (d+1) integer relations matrix with bit size b and test
```
malb: More whitespace changes.

tests/test_hnf.cpp
```diff
@@ -0,0 +1,144 @@
+/* Copyright (C) 2018 Arnaud Sipasseuth (inspired from the test_lll.cpp)
```
malb: We should have some known answer tests, e.g. comparing the HNF we get with that output by Sage or Flint.

malb:
Sorry for being so pedantic, but in my experience if we don't fix up this kind of stuff when it goes in, it is never fixed up.

malb:
You also wrote "This is not the final version, don't use hnf on large matrices or large determinants (it's slow)." Do you have an estimate when the final version would be roughly available? I'm asking because I'm hesitant to ship non-final code but we might have to if we merge this and the final version then takes a while (it can be done, just want to know). Also, how slow/fast is it? Are we talking dimension 10 or 100?

fplll/defs.h
arnsip: Yup you're right, it is wrong.

fplll/defs.h
arnsip: It's true, I used 6 like the size of LLL_METHOD_STR which has size 6 too while it only had 4 enums defined above. I am not sure why to be honest I should have asked before, but I'll fix that to 3 while it might grow up later.

fplll/hnf.cpp
arnsip: I tried to look similar to others but I will remove that.

fplll/hnf.cpp
arnsip: It actually look both like copy-pasted code because it actually is for most of it. I didn't want to make a second loop for the reduction part that detects at the same time if there's a break on the diagonal form (not all HNF have a strictly diagonal form, it also depends of rank and sizes), so I included the reduction part in the detection all together so I can have only a single loop and not have another diagonal pivot detection to run when reducing the copy of the vector/matrix for checking if it is in the lattice.

fplll/hnf.h
arnsip: Because this only works if B is given is a HNF, but my naming is indeed confusing. I didn't want to include the HNF computation part. I will rename this to in_lattice_given_HNF. Makes things clearer.

fplll/latticegen.cpp
arnsip: ...I guess I haven't finished cleaning the previous mess.

fplll/hnf.h
malb: How expensive is it to test `is_hnf_reduced`?

fplll/nr/matrix.h
arnsip: I don't think so either, but when I saw it actually returns an error message when the matrix is not square, I thought there was a specific reason for it and decided to create a new one to be sure I don't potentially break other parts of the code.

I can merge them both in the same function "gen_uniform" indeed.

fplll/hnf.cpp
malb: Maybe the common code could be factored out into a separate helper function which is then called by the higher level functions?

tests/test_hnf.cpp
arnsip: Compared to FLINT (and MAGMA), I actually create lower diagonal HNF, instead of upper diagonal ones. This is to be consistent with the option "T d" on latticegen which, in that case, generate a random matrix that is almost a HNF.

However, we can still verify using FLINT or magma : one has to use a central symmetry before the computation and then after the computation, and check if both results match. It should match (in my small example cases it matched).

tests/Makefile.am
arnsip: I'm not sure if it's if the editor or my previous screw-up with clang-format. :s

fplll/hnf.cpp
arnsip: I'll try that.

fplll/nr/nr_Z.inl
arnsip: For now I've limited myself to mpz_t for HNF computations. The main function should return an error when tried to force a non-gmp integer type for HNF computations. I can try to code those number theoretical function for all data types, it will take some extra time. To be fair I will probably add some for modular computations.

I added those in the integer class for the sole purpose to make the HNF code more readible, which isn't templated (yet?).

malb:
Okay, let's hold off with merging this for the time being then but I'll continue to provide feedback. Do let me know when that becomes frustrating, though, as we really appreciate your efforts! Do you think it makes to add a `HNF` object similar to the `BKZ` objects for API symmetry?

arnsip:
I guess it's not a bad idea to make a HNF object for API symmetry, I am more used to a C-style coding (without objects) but ultimately once we get all functions it doesn't hurt to have classes for the sake of consistency. One week was a bit optimistic: it's better to hold off.

malb:
Closing this, this has been inactive for a while

PR #346, opened 2018-01-24
minor fix sieve outputs and template inst.

shi-bai:
Current sieve does not output the shortest vector found in the end. Just added that.

Also add some other instantiation declaration for other float types in case the users wanted to call them.

### DIFF
--- a/fplll/sieve/sampler_basic.cpp
+++ b/fplll/sieve/sampler_basic.cpp
@@ -155,7 +155,9 @@ template <class ZT, class F> NumVect<Z_NR<ZT>> KleinSampler<ZT, F>::sample()
 
 template class KleinSampler<long, FP_NR<double>>;
 template class KleinSampler<mpz_t, FP_NR<double>>;
+#ifdef FPLLL_WITH_QD
 template class KleinSampler<long, FP_NR<dd_real>>;
 template class KleinSampler<mpz_t, FP_NR<dd_real>>;
+#endif
 template class KleinSampler<long, FP_NR<mpfr_t>>;
 template class KleinSampler<mpz_t, FP_NR<mpfr_t>>;
--- a/fplll/sieve/sampler_basic.cpp
+++ b/fplll/sieve/sampler_basic.cpp
@@ -155,4 +155,7 @@ template <class ZT, class F> NumVect<Z_NR<ZT>> KleinSampler<ZT, F>::sample()
 
 template class KleinSampler<long, FP_NR<double>>;
 template class KleinSampler<mpz_t, FP_NR<double>>;
+template class KleinSampler<long, FP_NR<dd_real>>;
+template class KleinSampler<mpz_t, FP_NR<dd_real>>;
+template class KleinSampler<long, FP_NR<mpfr_t>>;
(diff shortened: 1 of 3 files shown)


fplll/sieve/sieve_gauss.cpp
```diff
@@ -292,16 +292,12 @@ template <class ZT, class F> void GaussSieve<ZT, F>::print_final_info()
...
     cout << "# [info] true log2(max|L|)/n = " << log2(first_size) / nc << endl;
-    cout << "# [info] sv is" << endl;
-  }
-  if (verbose)
-  {
-    final_norm.set_z(best_sqr_norm);
-    final_norm.sqrt(final_norm, GMP_RNDN);
-    cout << "# [info] |sv| = " << final_norm << " (" << best_sqr_norm << ")" << endl;
   }
+  cout << "# [info] sv is " << return_first() << endl;
```
malb: I have to admit, I'm not too keen on the unconditional printing here. For example, if this code is used inside fpylll then there's a bunch of spurious printed messages when calling this function in a loop.

shi-bai: Right, I remembered that now. I added it since current ./latsieve binary will not output the vector found which is a bit strange.  Let me think what to do with that.

shi-bai: Martin, now I enabled the verbose = True when called from command line which will output the shortest vector. When used from fpylll, one can use verbose = False to GaussSieve object to avoid the verbose problem.

shi-bai:
Strange that my Jan 24 pull request passed all checks; but today it reports 

> sieve/sieve_gauss.cpp:322:39: error: use of undeclared identifier 'dd_real' template class GaussSieve<long, FP_NR<dd_real>>;

on the 
Apple LLVM version 8.1.0 (clang-802.0.42)
Target: x86_64-apple-darwin16.7.0

I am lost. Any suggestions?

fplll/sieve/sieve_gauss.cpp
```diff
@@ -320,7 +317,11 @@ template <class ZT, class F> bool GaussSieve<ZT, F>::sieve(Z_NR<ZT> target_norm)
     return run_2sieve();
 }
 
-template class GaussSieve<mpz_t, FP_NR<double>>;
 template class GaussSieve<long, FP_NR<double>>;
+template class GaussSieve<mpz_t, FP_NR<double>>;
+template class GaussSieve<long, FP_NR<dd_real>>;
```
malb: You are missing `#ifdef` guards around the dd/qd stuff.

fplll/sieve/sampler_basic.cpp
```diff
@@ -155,4 +155,7 @@ template <class ZT, class F> NumVect<Z_NR<ZT>> KleinSampler<ZT, F>::sample()
 
 template class KleinSampler<long, FP_NR<double>>;
 template class KleinSampler<mpz_t, FP_NR<double>>;
+template class KleinSampler<long, FP_NR<dd_real>>;
```
malb: You are missing `#ifdef FPLLL_WITH_QD`

PR #348, opened 2018-02-05
Fix Solaris issues

jdemeyer:
This fixes two issues when compiling on Solaris with GCC-5.4.0.

The first is an issue with `pow()`:
```
nr/dpe.h: In function 'void dpe_pow_si(dpe_struct*, const dpe_struct*, unsigned int)':
nr/dpe.h:718:35: error: call of overloaded 'pow(const double&, const unsigned int&)' is ambiguous
   DPE_MANT(x) = pow(DPE_MANT(y), k);
```

The second is GCC bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=10200
The existence of a templated `norm` function causes GCC to misparse `if (p->norm < v->norm)`


malb:
Thank you!

PR #349, opened 2018-02-07
Possibly resolved the Gram bug

kodebro:
Resolved the Gram-matrix bug by replacing `n_known_rows` in the Gram-operations by the full dimension `d`. 

This makes sense, because only a partial operation [up tp `n_known_rows`] on the Gram-matrix is not consistent and causes irregularities. In larger dimensions, this cause distortion of the positive-definiteness, causing severe errors during LLL-reduction.

This particular code also occurs in gso.cpp. We could also replace `n_known_rows` by `d` there. Is there a reason that it is `n_known_rows`? **Edit: yes. If `enable_int_gram` the exact Gram matrix is stored, but only up to `n_known_rows`, therefore that this code should stay in gso.cpp.**

### DIFF
--- a/fplll/gso_gram.cpp
+++ b/fplll/gso_gram.cpp
@@ -68,7 +68,7 @@ template <class ZT, class FT> void MatGSOGram<ZT, FT>::row_add(int i, int j)
     ztmp1.add(ztmp1, g(j, j));
     g(i, i).add(g(i, i), ztmp1);
 
-    for (int k = 0; k < n_known_rows; k++)
+    for (int k = 0; k < d; k++)
     {
       if (k != i)
         sym_g(i, k).add(sym_g(i, k), sym_g(j, k));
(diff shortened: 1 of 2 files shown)


lducas:
:+1:

malb:
I don't know the code enough to make a judgement call on `n_known_rows` vs `d`, @lducas is your 👍 your review?

kodebro:
@malb `n_known_rows` is the number of rows that the GSO-module was able to gram-schmidt reduce; we saw examples of matrices with big coefficients where gram-schmidt stopped after a few rows [say, row 10 out of 100]. Then d = 100, but n_known_rows = 10. 
In  small examples, the gram-schmidt succeeds up to the dimension `d`, so that `d` equals `n_known_rows`. Then the old code succeeds, because those two variables are equal.
In the case where we got `n_known_rows` < `d`, [so the gram-schmidt stopped somewhere], we got errors, because the Grammatrix was only partially modified.

malb:
Thanks, that helps already. How can we compute with coefficients, though, that the GSO module couldn't compute? Or are they fine, they are simply not reduced yet? Sorry if that's daft.

kodebro:
No problem!
The Gram coefficients are given! Because the input is a [full] Gram matrix. So, the basis operation on this Gram matrix must be 'fully' applied.

PR #354, opened 2018-05-19
pruner using full coefficients and optimization routines with fixed probability

shi-bai:


shi-bai:
Thanks Leo :) Indeed the pruner.cpp/.h are split into several smaller files for easier management (hopefully). The functions in pruner.cpp/.h are now in in several files:

1. pruner_simplex.h (inline functions -- not changed at all)

2. pruner_cost.cpp (functions for computing the enumeration cost. Since now the code uses full length coefficients, the single_enum_cost() calls single_enum_cost_lower() and single_enum_cost_upper() and extrapolate the overall cost.) 

3. pruner_prob.cpp (the structure is very similar to pruner_cost.cpp to cater for the full length coefficients.)

Above are the basic functions. The main optimization interfaces are in pruner_optimize.cpp. The specific optimization routines are in pruner_optimize_tc.cpp and pruner_optimize_tp.cpp. The first one is relevant to optimize the overall cost. The second one is to optimize the single_enumeration_cost fixing the probability.

4a) optimize_coefficients_cost() is the main function to optimize the overall cost: single_enumeration_cost()/probability.   It optimizes the half-coefficient first; and then optimize the full-coefficients in several iterations. Within each iteration, it does some tunning and do  optimize_coefficients_full() (this does the gradient/Nelder-Mead as before).

4b) optimize_coefficients_prob() is the main function to optimize the single enumeration cost while fixing probability. It does a small full optimization from 4a) and then adjust the coefficient to achieve the probability (while hopefully maintaining the curve shape of the pruning coefficients.)

The current routines are slower than before but hopefully the quality could be better.

Cheers,
Shi

fplll/bkz.cpp
```diff
@@ -803,11 +803,107 @@ template <class ZT, class FT> bool BKZAutoAbort<ZT, FT>::test_abort(double scale
   return no_dec >= maxNoDec;
 }
 
+// call pruner
+template <class FT>
+int bkz_prune_f(ZZ_mat<mpz_t> &b, const BKZParam &param, int sel_ft, double lll_delta,
```
malb: I don't see why this function is called `bkz_` or why it is in `bkz.cpp`, it just calls the pruner, right?

fplll/Makefile.am
```diff
@@ -81,7 +81,7 @@ libfplll_la_SOURCES=fplll.cpp fplll.h \
 	bkz.cpp bkz.h \
 	bkz_param.cpp bkz_param.h \
 	gso_interface.cpp gso_interface.h gso_gram.cpp gso_gram.h gso.cpp gso.h \
-	pruner.cpp pruner.h \
+	pruner/pruner.cpp pruner/pruner.h \
```
malb: Looks like doctests fail because you forgot some header files here.

fplll/bkz_param.h
```diff
@@ -91,6 +91,7 @@ class BKZParam
           - BKZ_GH_BND        use the Gaussian heuristic to reduce the enumeration bound of possible
           - BKZ_SD_VARIANT    run SD-BKZ
           - BKZ_SLD_RED       run slide reduction
+          - BKZ_PRUNE_ONLY    only run pruner
```
malb: Same as above, I don't see how this is a `BKZ_` thing.

fplll/bkz_param.h
```diff
@@ -113,12 +114,14 @@ class BKZParam
            int auto_abort_max_no_dec      = BKZ_DEF_AUTO_ABORT_MAX_NO_DEC,
            double gh_factor               = BKZ_DEF_AUTO_ABORT_MAX_NO_DEC,
            double min_success_probability = BKZ_DEF_MIN_SUCCESS_PROBABILITY,
-           int rerandomization_density    = BKZ_DEF_RERANDOMIZATION_DENSITY)
+           int rerandomization_density = BKZ_DEF_RERANDOMIZATION_DENSITY, int prune_start = 0,
+           int prune_end = 0, double prune_pre_nodes = 1e8, int prune_min_prob = -1)
```
malb: Same as above.

fplll/bkz.cpp
shi-bai: Yes Martin. One reason is that the pruner can do a BKZ-like task: optimizing the pruning coefficients for block [i, j]. Also I want to re-use the bkz_reduction() function which filters the floating point types (to avoid to have another function do the same task).

malb: I'd rather say that BKZ also does a Pruning task, so I wouldn't put the pruning stuff in BKZ :) For the second reason, I'd re-factor and move the FP filtering into its own subfunction, that can be re-used

shi-bai: That sounds a good solution and we could move the bkz_prune_f to pruner.cpp. Let me take care of this.

malb: Ta!

fplll/Makefile.am
shi-bai: Yes ... I will fix this tomorrow.

fplll/pruner/pruner.h
```diff
@@ -0,0 +1,709 @@
...
+    }
+    load_basis_shapes(gso_rs);
+  }
+
+  /**
+     Run the optimization process, successively using the algorithm activated
+
+     @brief run the optimization process
+  */
+  void optimize_coefficients_evec(/*io*/ vector<double> &pr);
```
lducas: It would be nice to understand the hierarchy and differences between all these functions.

fplll/pruner/pruner_optimize.cpp
```diff
@@ -0,0 +1,152 @@
...
+  cerr << "# [stage] half-optimization done." << endl;
+  cerr << "# [stage]  all_enum_cost    is " << cost << endl;
+  cerr << "# [stage]  svp_probability is " << prob << endl;
+#endif
+}
+
+template <class FT> void Pruner<FT>::optimize_coefficients(/*io*/ vector<double> &pr)
+{
+
+  if (opt_overall)
```
lducas: Is this condition useful

shi-bai:
Hi Leo,
> your PR seems to add two feature:
> 1/ removed restriction to ``even'' vectors
Let me refer to this as "using full coefficients" in the later context.

> 2/ mechanism to enforce a given success proba.
Yes.
 
> I'm not convinced of the usefulness of 1/. Have you found instances were you ended up with a significantly better running time thanks to it.
As in the other email, I think if one fixes the pruning coefficients, then "using full coefficients" does not bring too much advantage in calculating the single enumeration time and probability (but still I can see 20% differences). But "using full coefficients" may also help in the optimization procedure. Perhaps one gets stuck at some local minimum using half coefficient. I attached some input lattice for SVP140. 

[svp140_intermediate.txt](https://github.com/fplll/fplll/files/2026595/svp140_intermediate.txt)

With radius 9841467.97515 and preprocess cost 1e8,

1) Using the pull request, I can get,
 ./fplll  -a pru -b 140 svp140_intermediate.txt  -f dd -bkzghbound 1 -pruminprob -1 -pruprenodes 1e8

# [Stage fio]  all_enum_cost    is 1.640807e+18
# [Stage fio]  single_enum_cost is 1.548507e+09
# [Stage fio]         succ_prob is 9.946457e-10

2) Using the current FPYLLL pruning with Pruning.ZEALOUS (here attached Python script).
[main_prune_clean.txt](https://github.com/fplll/fplll/files/2026603/main_prune_clean.txt)

I got 
# single_enum_cost =  68076059619.6
# metric =  solutions
# succ.prob =  8.69889096934e-09
# overall_cost =  7.82583203532e+18

The overall cost is improved (from 7.8e18 to 1.6e18). I also attached the picture for the pruning parameter comparison (the red dots are from FPYLLL; the blue are form pull request)
svp140_compare_prune

> For 2/:
> I could not identify how the API changed to choose or not this new option of forcing the success proba. I'm also not sure to understand exactly the new descent algorithm. Basically run the usual descent entirely, and then tweak parameters to force proba, and restart the descent ? I see no reason this should converge.

That's right. Using global optimization + tweaking probability seems to be very useful in practice (as the examples compared with pbkz shown in our other email)

> The problem of restrained optimization is indeed tricky. I think that we should re-enforce the restriction at each step of the descent, not only at the end of it.
> 
> Maybe I'm being too defensive about my pruner. But I feel like requesting to restart this whole process from scratch. Before we decide how to code those: what features do you want and why ? Please understand that I really want to maintain so form of minimality for the pruner, because it is very hard to test that it does what it is supposed to do.
No problem. That's actually part of the reason that I didn't touch the previous gradient code at all. I only used them as subroutines. I hope somehow that maintains the integrity from previous versions.

shi-bai:
Leo,

Here is another example on smaller lattices. Input is SVP120 with BKZ45 reduced.

Radius is 8433658 and preprocessing nodes 1e8,

Pull request gives,
 [Stage fio] final optimization done.
 [Stage fio]  all_enum_cost    is 4.813470e+15
 [Stage fio]  single_enum_cost is 1.126811e+09
 [Stage fio]         succ_prob is 2.523218e-07


FPYLLL pruning with Pruning.ZEALOUS gives,
 single_enum_cost =  1004512103.92
 metric =  solutions
 succ.prob =  1.76261789944e-07
 overall_cost =  5.69897823144e+15

It goes form 5.7e15 to 4.8e15.

Images are here,

svp120_compare_prune

fplll/pruner/pruner.h
shi-bai: Yep, updated.

lducas:
Ok, lets keep feature 1 for later. First I would like to get feature 2 done properly. Maybe we can start another branch for that, just to test if what I suggest is, as I hope, simpler and more efficient.

There is this function ``enforce'' that is called very often: it is used to make sure that the visited points stays in the domain of interest. If we want to force success probability to a given value, I think this should be added in this function, with a global option. I suspect the results would be much better (better minimum), or at least much faster (faster convergence).

Can you give it a shot ?

Best

shi-bai:
Leo,

> There is this function ``enforce'' that is called very often: it is used to make sure that the visited points stays in the domain of interest. If we want to force success probability to a given value, I think this should be added in this function, with a global option. I suspect the results would be much better (better minimum), or at least much faster (faster convergence).

Yes, there is the single_enum_cost_gradient() function which was thought useful to do such task in descent. It turned out the descent (or heuristicly dec/inc coefficients) part is fine; but figuring out a good "step amount" is the hard part. I mean the following case: say we want targeted succ. prob 1e-4. The current coefficient say gives 1e-3. We want to decrease the coefficients to get 1e-4 from 1e-3. The main issue is we may need to trial-error to test several step amounts. If the step is too large, it may drop well-below 1e-4. This turns out to be pretty sensitive and we may use a binary search to determine the step. So I thought we can't easily adapt the current gradient_descent() for this. I agree this would be a better approach but we may need independent functions for that. Updated a working branch for this. 

Cheers,
Shi

lducas:
Hi Shi,

I looked at the new branch, and the documentation looks better, though I have not went into the details. 

I must admit I'm growing tired of the pruner. I suggest the following way to handle this PR:

- the previous behaviour of the pruner remain available with an option (but if the new one behave better, the default option should be the new one).
- A bit more level is done on documentation. In particular I'm still unsure what is done to handle ``full'' instead of ``half'', and what is precisely optimized.
- You will now on be responsible for the pruner ;)

fplll/bkz.cpp
shi-bai: The command line stuff is factored out of bkz.cpp into pruner.cpp. The bkz.cpp and bkz.h are not touched in the pull request now.

fplll/bkz_param.h
shi-bai: Fixed.

fplll/bkz_param.h
shi-bai: Fixed.

fplll/Makefile.am
shi-bai: I think this is fixed?

fplll/pruner/pruner.h
```diff
@@ -0,0 +1,709 @@
...
+  // void greedy_full(evec &b);
+
+  /**
+     @brief Perform one step of the gradient descent, in place
+     @param b input/output
+  */
+  int gradient_descent_step(/*io*/ vec &b);
+
+  int gradient_descent(/*io*/ vec &b);
+
```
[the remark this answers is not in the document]
shi-bai: Fixed.

fplll/pruner/pruner_optimize.cpp
shi-bai: If (opt_overall), it will optimize the overall cost = single_enum_cost / succ. probability. If not, it will optimize the single_enum_cost by fixing the input target succ. probability.

fplll/pruner/pruner.cpp
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  adjust_radius_to_gh_bound(max_dist, max_dist_expo, block_size, root_det, param.gh_factor);
+  double radius_d = max_dist.get_d() * pow(2, max_dist_expo);
+
+  cerr << "# Start Pruning" << endl;
+  cerr << "# enumeration Radius: " << radius_d << endl;
+  cerr << "# preprocessing (num. nodes): " << prune_pre_nodes << endl;
+  cerr << "# targeted min. prob: " << prune_min_prob << endl;
+  cerr << "# input GSO: " << r << endl;
```
malb: This looks quite noisy.

malb:
If I read the code correctly, this PR changes the semantics of `PruningParams`. Right now, they are purely outputs of the pruner but not inputs. But in these objects are also used as inputs. I don't think that's a good design decision.

fplll/pruner/pruner.cpp
shi-bai: These are outputs from the pruner. They look informative to me. Perhaps we can remove the last one. 

cerr << "# input GSO: " << r << endl;

shi-bai:
> If I read the code correctly, this PR changes the semantics of `PruningParams`. Right now, they are purely outputs of the pruner but not inputs. But in these objects are also used as inputs. I don't think that's a good design decision. 

Yes indeed. Alternative we could do like what lll_reduction() does by just passing the input parameters as function parameters?

fplll/pruner/pruner.cpp
malb: But doesn't that mean we cannot call the pruner without stderr being poluted with debug information? That is quite noisy, e.g. think in a multicore setting.

malb:
I think turning those into function parameters is probably the best option.

fplll/pruner/pruner.cpp
shi-bai: The above output is only for the fplll command line usage. If someone uses ./fplll -pru in the command line, they would expect some outputs. At least the pruning parameter should be output otherwise they got no results. 

I am not sure if fplll has a multicore support. Do you mean the call from fpylll? Then the behavior should be the same as the current fpylll (there is no output except writing the result to the bkzparam).

shi-bai:
Let me do that.

fplll/pruner/pruner.cpp
malb: Ah! Sorry, my bad. All good on this one then.

malb:
Once doctests pass and highlevel questions are settled, I'll write an fpylll interface to test this code, this should reveal any outstanding interface kinks.

malb:
FWIW, I think those coverage failures are acceptable. We're more careful about what to build which means some stuff isn't covered.

fplll/defs.h
```diff
@@ -286,6 +286,8 @@ enum PrunerFlags
   PRUNER_NELDER_MEAD      = 0x8,
   // Verbosity
   PRUNER_VERBOSE = 0x10,
+  // Optimize w.r.t to all coefficients (by default it optimizes w.r.t half)
+  PRUNER_OPTIMIZE_FULL = 0x20
```
malb: Maybe just `PRUNER_FULL` since `OPTIMIZE` is kinda redundant?

fplll/pruner/pruner.h
```diff
@@ -16,7 +13,11 @@
    You should have received a copy of the GNU Lesser General Public License
    along with fplll. If not, see <http://www.gnu.org/licenses/>. */
 
+#ifndef FPLLL_PRUNER_H
+#define FPLLL_PRUNER_H
+
 #include "defs.h"
+#include "lll.h"
```
malb: Those includes should be for "fplll/defs.h" and "fplll/lll.h", otherwise third party code like FPyLLL  won't build.

fplll/pruner/pruner.h
```diff
@@ -124,10 +188,18 @@ template <class FT>
 void prune(/*(input)output*/ PruningParams &pruning,
            /*inputs*/
            const double enumeration_radius, const double preproc_cost, const vector<double> &gso_r,
-           const double target       = .9,
+           const double input_target = .9,
```
malb: Why `input_target`?

fplll/defs.h
shi-bai: Yep, did the succinct one.

fplll/pruner/pruner.h
shi-bai: Thanks. I changed them. Let me know if there are other header problems.

fplll/pruner/pruner.h
shi-bai: Changed back to target. The purpose was: the input_target could be changed to something else in this->target after some sanity check.

fplll/pruner/pruner.h
```diff
@@ -128,6 +192,14 @@ void prune(/*(input)output*/ PruningParams &pruning,
            const PrunerMetric metric = PRUNER_METRIC_PROBABILITY_OF_SHORTEST,
            const int flags           = PRUNER_GRADIENT);
 
+template <class FT>
+double prune_cost(/*(input)output*/ PruningParams &pruning,
```
malb: This function doesn't have documentation. Also, AFAIK `pruning` is not an output for this function?

fplll/pruner/pruner.h
```diff
@@ -290,47 +390,207 @@ template <class FT> class Pruner
...
 
   /**
+     @brief run the optimization process using 'even' coefficients. Note
+     the optimization only applies to the pruning coefficients indexed
+     by (0, 2, 4, ... n). It thus uses half of the coefficients.
+
+     Run the optimization process, successively using the algorithm activated
+     using using half coefficients.
+  */
+  void optimize_coefficients_evec(/*io*/ vector<double> &pr);
```
malb: Is the documentation up to date, i.e. is pr supposed to be a vector of length `n/2` or of length `n`? The `Pruner` class gets a flag for half vs full, so I don't see why there's an additional choice of public function to select which one to use?

fplll/pruner/pruner.h
```diff
@@ -290,47 +390,207 @@ template <class FT> class Pruner
...
-  void optimize_coefficients(/*io*/ vector<double> &pr);
+  void optimize_coefficients_full(/*io*/ vector<double> &pr);
+
+  /**
+     @brief tune the pruning parameter to reduce single enumeration time
+
+     Optimization process to the pruning parameters to reduce single enumeration
+     time with the hope that it also reduces the overall enumeration time.
+  */
+  void optimize_coefficients_tune_cost(/*io*/ vector<double> &pr);
```
malb: This should perhaps be called `tune_single`?

fplll/pruner/pruner.h
```diff
@@ -290,47 +390,207 @@ template <class FT> class Pruner
...
+  void optimize_coefficients_smooth(/*io*/ vector<double> &pr);
+
+  /**
+     @brief main interface to optimize the overall enumeraiton time
+
+     Main interface to optimize the overall enumeraiton time where the
+     target unciton is: single_enumeration_cost divided by succ. probability
+     or expected solutions.
+  */
+  void optimize_coefficients_cost(/*io*/ vector<double> &pr);
```
malb: This doesn't seem correct, the overall cost is (t-1)*preproc + t*enum where t is roughly 1/pr, no?

fplll/pruner/pruner.h
```diff
@@ -290,47 +390,207 @@ template <class FT> class Pruner
...
+     computed using pruning parameters. Then one decreases the succ.
+     probability by descreasing the pruning parameters.
+  */
+  void optimize_coefficients_prob_decr(/*io*/ vector<double> &pr);
+
+  /**
+     @brief auxiliary function in optimizing the single enumeraiton time fixing succ. prob.
 
-  /** @brief Compute the cost of a single enumeration */
+     Heuristic tuning procedure which seems to be useful.
```
malb: This is really vague, can we say more?

fplll/pruner/pruner.h
```diff
@@ -290,47 +390,207 @@ template <class FT> class Pruner
...
+
+  /**
+     @brief Main interface to optimize pruning coefficients
+
+     Main interface to optimize pruning coefficients. It will invoke either:
+     optimize_coefficients_cost() or
+     optimize_coefficients_prob()
+     depending on the input "target". If the target is negative (e.g. -1),
+     it calls the first function where the goal is to optimize the
+     single_enum_cost() divided by the succ. probability. If the target
```
malb: See above

fplll/pruner/pruner.h
```diff
@@ -290,47 +390,207 @@ template <class FT> class Pruner
...
+     is > 0, it calls the second function where the goal is to optimize the
+     single_enum_cost() while fixing the succ. probability == target.
```
malb: second function?

fplll/pruner/pruner.h
```diff
@@ -290,47 +390,207 @@ template <class FT> class Pruner
...
+  */
+  void optimize_coefficients_prob_tune(/*io*/ vector<double> &pr);
+
+  /**
+     @brief Main interface to optimize pruning coefficients
+
+     Main interface to optimize pruning coefficients. It will invoke either:
+     optimize_coefficients_cost() or
+     optimize_coefficients_prob()
+     depending on the input "target". If the target is negative (e.g. -1),
```
malb: Why overload target like that, isn't this better handled by a flag?

fplll/pruner/pruner.h
```diff
@@ -290,47 +390,207 @@ template <class FT> class Pruner
...
+     Compute the cost of r enumeration and (r-1) preprocessing, where r is the
+     required number of retrials to reach target.
+
+     There are two modes depending on the target.
+
+     (1) if target == -1, we will optimize the single_enum_cost/prob.
+
+     (2) if target != -1 (target must also be given),  we will optimize
+     the single_enum_cost while fixing the input target prob. In such case,
+     the repeated_enum_cost() actually returns single_enum_cost() since it
```
malb: I think the function should be renamed in this case

fplll/pruner/pruner.h
```diff
@@ -290,47 +390,207 @@ template <class FT> class Pruner
...
 
-     Compute the cost of r enumeration and (r-1) preprocessing, where r is the required number of
-     retrials to reach target.
+     Compute the cost of r enumeration and (r-1) preprocessing, where r is the
+     required number of retrials to reach target.
+
+     There are two modes depending on the target.
+
+     (1) if target == -1, we will optimize the single_enum_cost/prob.
+
```
malb: This function doesn't optimize (?)

fplll/pruner/pruner.h
```diff
@@ -407,6 +675,26 @@ template <class FT> class Pruner
...
 
+  /**
+     @brief convert pruning coefficient from external to internal format.
+
+     Convert type, reverse the ordering, and select all coefficents
+
+     @param fb Output in internal format
+     @param pr Input in external format
+  */
+  void load_full_coefficients(/*o*/ vec &fb, /*i*/ const vector<double> &pr);
```
malb: Looks like these are conversion and not loading functions(?)

malb:
Hi, FPyLLL now compiles against this branch. I've left a billion style comments. Sorry about that, but to me, the API is the main user interface and in my experience, if we don't make it nice now, we'll never get around to doing it.

fplll/pruner/pruner.h
shi-bai: My bad. This function is not used anywhere. I will remote it.

fplll/pruner/pruner.h
shi-bai: I updated it. The input pr is of length n; but the function optimize_coefficients_evec(pr) will only use the even indices among them (this is to be compatible with previous version). 

Yes, the flag to the Pruner class decides whether it is full or half. If it is half, it will call optimize_coefficients_evec(pr).

fplll/pruner/pruner.h
shi-bai: Done.

fplll/pruner/pruner.h
shi-bai: Added

fplll/pruner/pruner.h
shi-bai: Yes, fixing now.

fplll/pruner/pruner.h
shi-bai: Fixing now

fplll/pruner/pruner.h
shi-bai: Fixed this.

fplll/pruner/pruner.h
shi-bai: Do you mean adding a flag "optimize_with_fixed_prob"  It will still need user to input a target prob. Seems a bit redundant for the extra flag?

fplll/pruner/pruner.h
shi-bai: Yes, this is a hack for the moment. The thing is we used it in the optimization procedure as the target function. But since we now have two types of targets, this function also changes accordingly. Perhaps we could make another function for such task? And only call this function in the end of the optimization to pass the values to PruningParams.

fplll/pruner/pruner.h
shi-bai: Added a comment yes.

fplll/pruner/pruner.h
shi-bai: Thanks for spotting. It's a ghost function not used ... removed it.

fplll/pruner/pruner.h
malb: Maybe have a private function for calling internally and a public function? I'd like to get rid of the hacks before merging since for me/FPyLLL the class API *is* the public interface.

shi-bai: I renamed the previous repeated_enum_cost() as target_function() (since it is what we are optimizing for: this could be a single_enum_cost()). Then there is a new repeated_enum_cost() which returns (single_enum_cost() + preprocessing) / succ. prob (or similar).

malb:
Thanks for these changes. I'm still a bit unclear about the API of the Pruner class. The master version of this class has essentially two modes: (a) optimise for finding the shortest vector or (b) optimise to find `x` vectors within the given radius. The mode is selected using a metric flag.

This PR extends this to: (a) optimise for the shortest vector, (b) optimise to find `x` vectors within the given radius and (c) mimise pruning cost of a single enumeration but with a fixed success probability (?) However, these three modes are partly selected by metric flag and partly by calling a different function. If my understanding is correct, then it's not clear to me why that's the case.

Additionally, this PR adds the option to select full or half coefficients. This can be selected both by a flag (at object creation) and by calling a method. Perhaps

```
  void optimize_coefficients_evec(/*io*/ vector<double> &pr);
  void optimize_coefficients_full(/*io*/ vector<double> &pr);
```

should be private? Also, shouldn't `_evec` be `_half` for consistency? Or am I missing something?

I'm also a bit confused about the pair

```
  void optimize_coefficients_cost(/*io*/ vector<double> &pr);
  void optimize_coefficients_prob(/*io*/ vector<double> &pr);
```

Shouldn't the latter be `optimize_coefficients_cost_fixed_prob`, i.e. it doesn't optimize the probability analogously to `optmize_coefficients_cost` (?)

fplll/pruner/pruner.h
```diff
@@ -290,37 +382,186 @@ template <class FT> class Pruner
...
+  void optimize_coefficients_tune_single(/*io*/ vector<double> &pr);
+
+  /**
+     @brief tune the pruning parameter to increase succ. probability
+
+     Optimization process to the pruning parameters to increase succ.
+     probability with the restriction that the single enumeration time does
+     not increase significantly.
+  */
+  void optimize_coefficients_tune_prob(/*io*/ vector<double> &pr);
```
malb: It is not clear to me why these got a `_tune_` in their name, don't they all tune? Also, is this a helper function used internally or is it at the same level as `optimize_coefficients`?

fplll/pruner/pruner.h
```diff
@@ -290,37 +382,186 @@ template <class FT> class Pruner
...
+
+  /**
+     @brief tune the pruning parameter to make the curve a more smooth.
+
+     Optimization process to the pruning parameters to make the curve
+     more smooth if the input has a large gap between consecutive
+     pruning parameters.
+  */
+  void optimize_coefficients_smooth(/*io*/ vector<double> &pr);
```
malb: Same as above, not clear if this is an internal, private helper function or a high level function from the documentation.

fplll/pruner/pruner.h
```diff
@@ -290,37 +382,186 @@ template <class FT> class Pruner
...
 
-  /** @brief Compute the cost of a single enumeration */
+  /**
+     @brief main interface to optimize the overall enumeraiton time
+
+     Main interface to optimize the overall enumeraiton time where the
+     target unciton is:
```
malb: -> function

fplll/pruner/pruner.h
```diff
@@ -290,37 +382,186 @@ template <class FT> class Pruner
...
+     @brief main interface to optimize the single enumeraiton time fixing succ. prob.
+
+     Main interface to optimize the single enumeration time with the constraint
+     that the succ. prob (or expected solutions) is fixed.
+  */
+  void optimize_coefficients_prob(/*io*/ vector<double> &pr);
+
+  /**
+     @brief auxiliary function in optimizing the single enumeraiton time
+     fixing the succ. prob.
```
malb: the `@brief` summary is the same for a few functions, making it more difficult to figure out their difference.

fplll/pruner/pruner.h
```diff
@@ -290,37 +382,186 @@ template <class FT> class Pruner
...
+     constraint that the succ. prob (or expected solutions) is fixed. It
+     is used if the given targeted probaility is larger than the one
+     computed using pruning parameters. Then one increases the succ.
+     probability by increasing the pruning parameters. Note since the
+     radius is fixed as input, it may not be possible to achieve the
+     given targeted probability even if all coefficients are (1, ..., 1).
+     In such case, it will try to increase the succ. prob as much as
+     it can and possibly return (1, ..., 1).
+  */
+  void optimize_coefficients_prob_incr(/*io*/ vector<double> &pr);
```
malb: Why not just `decrease_probability` or `decrease_prob` (to say with the abbreviation `prob` that's used throughout).

fplll/pruner/pruner.h
```diff
@@ -290,37 +382,186 @@ template <class FT> class Pruner
...
+
+  /**
+     @brief auxiliary function in optimizing the single enumeraiton time fixing succ. prob.
+
+     Heuristic tuning procedure which seems to be useful. This is used to make the ratio
+     between the succ. prob and the target succ. prob are close. Depending on whether
+     the succ. prob is larger (or smaller), it will try to reduce the pruning coefficients
+     (or increase) to make succ. prob \approx the target succ. prob. Sometimes the succ.
+     prob can not be modified closer to target succ. prob due to the contraints.
+     In such case it exists.
```
malb: exits? But perhaps `returns` is better, i.e. this function doesn't call `exit()`

fplll/pruner/pruner.h
```diff
@@ -290,37 +382,186 @@ template <class FT> class Pruner
...
+  */
+  void optimize_coefficients_prob_tune(/*io*/ vector<double> &pr);
```
malb: If you accept my naming suggesting above, then perhaps `adjust_prob` would wok here?

fplll/pruner/pruner.h
```diff
@@ -290,37 +382,186 @@ template <class FT> class Pruner
...
+     Main interface to optimize pruning coefficients.
+     It will in turn invoke either of the two functions:
+      (1)   optimize_coefficients_cost() or
+      (2)   optimize_coefficients_prob()
+     depending on the input "target".
+
+     If the target is negative (e.g. -1), it calls function (1)
+     such that goal is to optimize the
+     single_enum_cost(pr) * trials + preproc_cost * (trials - 1.0).
+
```
malb: I'm still not a fan of overloading `target` (given to the constructor) like this. The decision of whether to optimize globally or a single enumeration could be handled by a flag e.g. `PRUNER_SINGLE` or so. Also, why break existing code by changing the previous default behaviour, i.e. global optimization is used when target < 0? In `master` the user supplies a `target > 0` and gets a global optimisation, in this PR the user gets a local optimisation. Is that really the more common target?

shi-bai:
> Thanks for these changes. I'm still a bit unclear about the API of the Pruner class. The master version of this class has essentially two modes: (a) optimise for finding the shortest vector or (b) optimise to find x vectors within the given radius. The mode is selected using a metric flag.
>  
> This PR extends this to: (a) optimise for the shortest vector, (b) optimise to find x vectors within the given radius and (c) mimise pruning cost of a single enumeration but with a fixed success probability (?) However, these three modes are partly selected by metric flag and partly by calling a different function. If my understanding is correct, then it's not clear to me why that's the case.

Yes, to me the radius is an input parameter by the user, as it is the target vector length we want to find. So given an input radius, preprocessing cost, the b_i^*, there are essentially two optimization tasks: which are the above (a) and (c). The task (a) is default as it used to be. The flag "opt_overall" will trigger the task (c).

 -- If the task is (a), then for compatibility (see discussion with Leo), by default the code uses the even vectors for optimization. If one wants to use the full optimization, one can use the flag to enable it.

-- If the task is (b), then the code will call full optimization.

I see your point on the flag opt_overall v.s. target. I will comment there.

> Additionally, this PR adds the option to select full or half coefficients. This can be selected both by a flag (at object creation) and by calling a method. Perhaps
> 
>   void optimize_coefficients_evec(/*io*/ vector<double> &pr);
>   void optimize_coefficients_full(/*io*/ vector<double> &pr);
> should be private? Also, shouldn't _evec be _half for consistency? Or am I missing something?
Yes, it can be renamed to optimize_coefficients_half. It was just for compatibility since it's the function inherits from the Master. Both of them could be private. Perhaps only the optimize_coefficients() needs to be public.

> I'm also a bit confused about the pair
> 
>   void optimize_coefficients_cost(/*io*/ vector<double> &pr);
>   void optimize_coefficients_prob(/*io*/ vector<double> &pr);
> Shouldn't the latter be optimize_coefficients_cost_fixed_prob, i.e. it doesn't optimize the probability analogously to optmize_coefficients_cost (?)
The reason I use "optimize_coefficients_cost" and "optimize_coefficients_prob" is to make easier to remember they correspond to two target functions. Perhaps we can rename:

optimize_coefficients_cost --> optimize_coefficients_cost_vary_prob
optimize_coefficients_prob --> optimize_coefficients_cost_fixed_prob

fplll/pruner/pruner.h
shi-bai: The current optimize_coefficients_cost() and optimize_coefficients_prob() are two main functions. optimize_coefficients_tune_prob() is a helper function used internally.  Do you mean to put all such functions private?

fplll/pruner/pruner.h
shi-bai: Do you mean the following behavior? If flag is False, we ignore the target and do the optmize_coefficients_cost(). If flag is True, take the target into account and call the optmize_coefficients_prob(). 

I am just not sure if we should pass the PRUNER_SINGLE in the same existing flag. As the current flags 
"PRUNER_CVP  PRUNER_START_FROM_INPUT  PRUNER_GRADIENT  PRUNER_NELDER_MEAD  PRUNER_VERBOSE"
seem to be defined for the optimization "strategies"; but not for the "goals" of optimization. 

Or we could have an extra flag in the Pruner constructor which may be more clear.

fplll/pruner/pruner.h
shi-bai: Thank. No problem with above name convention, I will change :)

fplll/pruner/pruner.h
malb: In general, I'm in favour of public functions for a library like this as it makes it more hackable. In this case, though, I find these functions confusing, so I'd indeed either make them private or give them names that somehow distinguish better. Perhaps this is only my ignorance talking, though.

fplll/pruner/pruner.h
malb: Aren't we using `target` either way? We got three cases:
(a) find shortest vector with probability `target` with minimal overall cost
(b) find `target` vectors within the given radius
(c) find shortest vector with probability `target` in a single enumeration 

If I understand correctly, you are currently switching between (a) and (c) by using positive and negative signs for `target`. Isn't then (c) a special case of (a) but where the strategy is to use only one enumeration?

shi-bai: I thought I am switching between (b) and (c). Here is my understanding, there are two optimization goals:

Goal 1. Overall cost optimization:

single_enum_cost(b) * trials + preproc_cost * (trials - 1.0);
where the trials is computed either using svp_probability() or expected_solutions()

In such case, the target will be ignored since we don't fix the svp_probability() or expected_solutions(). We just want the overall cost to be small; the svp_probability() or expected_solutions()  may vary. I think this correspond to the case (a) above if using svp_probability().

Goal 2. Single enumeration cost optimization fixing probability/expectation.

single_enum_cost(b) while fixing `target' from input. 

In such case, we will try to achieve the `target' probability (or expected solutions) while reducing the single_enum_cost(). Still it may not guaranteed we can achieve this since the two inputs `target' and `radius' could be contradictory to each other. E.g. one set radius smaller than the GH and ask `target' very large. In these cases, it will try it's best to achieve the `target'. 

For either of the above 2 goals, there are two benchmarks:

One can use either svp_probability() or expected_solutions(). If one uses svp_probability(), then the input `radius' may not have any affects. If one uses expected_solutions(), the input `radius' may have some affects.

fplll/pruner/pruner.h
shi-bai: I made those "tune" functions private. I think it's an English question... I use "tune" to mean "small changes". There are other functions which does optimization but changes "bigger" on the pruning coefficients. I changed their name to "local_tune" and made them private --> let me know if this brings any problem to fpylll.

fplll/pruner/pruner.h
shi-bai: Changed similarly as mentioned above.

fplll/pruner/pruner.h
shi-bai: Yes! :)

fplll/pruner/pruner.h
shi-bai: This is done :) Now the optimization with fixed probability is triggered by the PRUNER_SINGLE with internal opt_single. This is compatible hopefully with current behavior in the master branch.

fplll/pruner/pruner.h
```diff
@@ -254,8 +316,42 @@ template <class FT> class Pruner
...
     set_tabulated_consts();
+
+    // optimize overall cost, use
+    if (flags & PRUNER_SINGLE)
+    {
+      opt_single = true;
+    }
+
+    // need to fix target if possible
+    if (metric == PRUNER_METRIC_PROBABILITY_OF_SHORTEST)
```
malb: Should we not throw an error here?

fplll/pruner/pruner.h
```diff
@@ -254,8 +316,42 @@ template <class FT> class Pruner
...
+      }
+    }
+    else if (metric == PRUNER_METRIC_EXPECTED_SOLUTIONS)
+    {
+      // check if the target is reasonable; otherwise set it to 0.99.
+      // note the target is allowed to be larger than 1 in this case.
+      // Also the target may affect the cost in target_function().
+      if (this->target < 0.0)
+      {
+        this->target = 0.99;
```
malb: Should we not throw an error here? Otherwise, I'd expect target = 1.0 to be a good choice?

fplll/pruner/pruner.h
```diff
@@ -290,37 +386,135 @@ template <class FT> class Pruner
...
+     @brief Main interface to optimize pruning coefficients
+
+     Main interface to optimize pruning coefficients.
+     It will in turn invoke either of the two functions:
+      (1)   optimize_coefficients_cost_vary_prob() or
+      (2)   optimize_coefficients_cost_fixed_prob()
+     depending on the input "target".
 
-     @brief run the optimization process
+     If the target is negative (e.g. -1), it calls function (1)
```
malb: I think this documentation needs updating

fplll/pruner/pruner.h
```diff
@@ -290,37 +386,135 @@ template <class FT> class Pruner
...
+
+     Run the optimization process, successively using the algorithm activated
+     using using half coefficients: the input pr has length n; but only the
+     even indices in the vector will be used in the optimization.
+     In the end, we have pr_i = pr_{i+1}.
+     Note it only optimize the overall enumeraiton time where the
+     target function is:
+     single_enum_cost(pr) * trials + preproc_cost * (trials - 1.0);
+  */
+  void optimize_coefficients_evec(/*io*/ vector<double> &pr);
```
malb: Maybe it is useful to explain the hierarchy of functions here: `optimize_coefficients` calls `optimize_coefficients_vary_prob` which calls `optimize_coefficients_evec`. Perhaps the latter should also be renamed to `optimize_coefficients_vary_prob_evec`?

fplll/pruner/pruner.h
```diff
@@ -462,49 +673,109 @@ template <class FT> class Pruner
...
+
+     (1) if target == -1, it will compute the cost of r * single_enum_cost()
+         plus (r-1) * preprocessing where r is the expected number of
+         trials.
+
+     (2) if target != -1 (target must also be given), it will compute
+         the cost of a single_enum_cost(). Note in such case the input
+         target prob is fixed. Whether we could achieve this input
+         target prob will be determined in the optimization procedure.
+         See also below.
```
malb: I think this documentation needs updating (?)

fplll/pruner/pruner.h
shi-bai: yes, it throws an exception now.

fplll/pruner/pruner.h
shi-bai: yep.

fplll/pruner/pruner.h
shi-bai: I added some explanation on the hierarchy of functions. The function optimize_coefficients_evec() is used in both optimize_coefficients_vary_prob() and optimize_coefficients_cost_fixed_prob() as a common function.

fplll/pruner/pruner.h
shi-bai: yes, updated now :)

fplll/pruner/pruner.h
```diff
@@ -290,47 +391,175 @@ template <class FT> class Pruner
...
+     depending on the input flags.
+
+     If the flag PRUNER_SINGLE is disabled (default),
+     it calls function (1) so that goal is to optimize the
+     single_enum_cost(pr) * trials + preproc_cost * (trials - 1.0).
+
+     If the flag PRUNER_SINGLE is enabled in `flags`,
+     it calls function (2) such that the goal is to optimize
+     the single_enum_cost(pr) while fixing
+     the succ. probability == target.
```
malb: I presume this should say "succ. probability or expected solutions == target"?

fplll/pruner/pruner.h
```diff
@@ -290,47 +391,158 @@ template <class FT> class Pruner
...
+
+     Run the optimization process, successively using the algorithm activated using using half
+     coefficients: the input pr has length n; but only the even indices in the vector will be used
+     in the optimization. In the end, we have pr_i = pr_{i+1}.
+
+     Note that this function only optimizes the overall enumeraiton time where the target function is:
+     `single_enum_cost(pr) * trials + preproc_cost * (trials - 1.0)`
+
+     This may be used in both `optimize_coefficients_cost_fixed_prob()` and
+     `optimize_coefficients_cost_vary_prob()`.
```
malb: So this function only does global optimization but may be called from `optimize_coefficients_cost_fixed_prob()`?

fplll/pruner/pruner.h
```diff
@@ -290,47 +391,158 @@ template <class FT> class Pruner
...
+     Hierarchy of calls: This function first invokes optimize_coefficients_evec() and then
+     optimize_coefficients_full() to optimize the overall enumeration cost. Then it tries to adjust
+     the pruning parameters to achieve the target succ. probability (or expected number of
+     solutions) by calling either optimize_coefficients_incr_prob() or
+     optimize_coefficients_decr_prob(). Finally, it does some local optimization by calling
+     optimize_coefficients_local_adjust_smooth() which aims to smooth the discountinuities in the
+     curve and then optimize_coefficients_local_adjust_prob() which aims to fine-adjust the succ.
+     probability to be close enough to the target.
+  */
+  void optimize_coefficients_cost_fixed_prob(/*io*/ vector<double> &pr);
```
malb: `optimize_coefficients_cost_fixed_prob` Sorry to be on the names again, but `prob` is only one metric, should this be `fixed_metric`, also should we drop the `_cost_`?

malb:
I've compiled `FPyLLL` against this branch and this now indeed works:

# Probability

## Global Optimisation

```python
FPLLL.set_random_seed(1337)
A = IntegerMatrix.random(40, "qary", bits=20, k=20)
_ = LLL.reduction(A)
M = GSO.Mat(A)
_ = M.update_gso()
pr = Pruning.Pruner(M.get_r(0,0), 2**10, [M.r()], 0.53)
c = pr.optimize_coefficients([1. for _ in range(M.d)])
pr.measure_metric(c)
```

0.22

## Local Optimisation

```python
FPLLL.set_random_seed(1337)
A = IntegerMatrix.random(40, "qary", bits=20, k=20)
_ = LLL.reduction(A)
M = GSO.Mat(A)
_ = M.update_gso()
pr = Pruning.Pruner(M.get_r(0,0), 2**10, [M.r()], 0.53, flags=Pruning.SINGLE)
c = pr.optimize_coefficients([1. for _ in range(M.d)])
pr.measure_metric(c)
```

0.5550643858111148

# Number of Solutions

When using the expected number of solutions as the metric, things behave unexpectedly (or perhaps this is to be expected?)

## Global Optimisation

```python
FPLLL.set_random_seed(1337)
A = IntegerMatrix.random(40, "qary", bits=20, k=20)
_ = LLL.reduction(A)
M = GSO.Mat(A)
_ = M.update_gso()
pr = Pruning.Pruner(M.get_r(0,0), 2**10, [M.r()], 1.0, metric=Pruning.EXPECTED_SOLUTIONS)
c = pr.optimize_coefficients([1. for _ in range(M.d)])
pr.measure_metric(c)
```
0.05416829905311384

## Local Optimisation

```python
FPLLL.set_random_seed(1337)
A = IntegerMatrix.random(40, "qary", bits=20, k=20)
_ = LLL.reduction(A)
M = GSO.Mat(A)
_ = M.update_gso()
pr = Pruning.Pruner(M.get_r(0,0), 2**10, [M.r()], 1.0, metric=Pruning.EXPECTED_SOLUTIONS, flags=Pruning.SINGLE)
c = pr.optimize_coefficients([1. for _ in range(M.d)])
pr.measure_metric(c)
```

0.22582101986062905

I'd have expected something close to 1.0 here.

shi-bai:
Thanks. I just tried the experiments in fplll. I got the same figures.

> I'd have expected something close to 1.0 here.

It seems the GH (which is 1.989655e+06) which is already larger than the current M.get_r(0,0) 1.912130e+06. The coefficients it returns are all 1's. The 0.22 is the expected solutions computed using the 1.912130e+06.

fplll/pruner/pruner.h
shi-bai: yes.

fplll/pruner/pruner.h
shi-bai: yes. it's used in both.

fplll/pruner/pruner.h
shi-bai: Maritn, I consider this more a constraint optimization problem, e.g.
optimize_coefficients_TargetFunction_Constraint

Currently we have 2 of them:
optimize_coefficients_cost_fixed_prob
optimize_coefficients_cost_fixed_prob

Presumably we may also have  optimize_coefficients_prob_fixed_cost in the future. In such case, the current naming convention seems easier to extend?

malb:
Ah, thanks re GH. This indeed works as expected:

```python
FPLLL.set_random_seed(1337)
A = IntegerMatrix.random(60, "qary", bits=30, k=30)
_ = LLL.reduction(A)
M = GSO.Mat(A)
_ = M.update_gso()
pr = Pruning.Pruner(0.9*M.get_r(0,0), 2**40, [M.r()], 1.0, metric=Pruning.EXPECTED_SOLUTIONS, flags=Pruning.SINGLE)
c = pr.optimize_coefficients([1. for _ in range(M.d)])
pr.measure_metric(c)
```

1.0490729740159583

However, I'm confused about:

```python
pr = Pruning.Pruner(0.9*M.get_r(0,0), 2**40, [M.r()], 1.0, metric=Pruning.EXPECTED_SOLUTIONS)
c = pr.optimize_coefficients([1. for _ in range(M.d)])
pr.measure_metric(c)
```

31.129138739682432

Why is this so big?

shi-bai:
> Why is this so big?

It seems that the 0.9*M.get_r(0,0) is much larger than the GH (about 2.47 times). Hence the number of solutions with is already pretty large with that bound. The pruner tries to reduce the pruning parameter to get a balance between single_enum_cost, preprocessing nodes and exp. solutions. I got the following pruning coefficients (my bound used is not exactly 0.9*M.get_r(0,0) but very close).

[1 1 0.786081 0.786081 0.629024 0.629024 0.486581 0.486581 0.435395 0.435395 0.394239 0.394239 0.356061 0.356061 0.320645 0.320645 0.287792 0.287792 0.257316 0.257316 0.229045 0.229045 0.202819 0.202819 0.184857 0.184857 0.167854 0.167854 0.146056 0.146056 0.136525 0.136525 0.122104 0.122104 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389 0.113389]

PR #370, opened 2019-05-25
Added solution for potential Win10 issue

tmmlaarhoven:
When trying to install fplll again in Windows 10, I had the minor issue mentioned above, and found a simple fix online -- it might be useful for others too if they encounter the same issue.


malb:
Ta!

PR #384, opened 2019-10-30
Extenum documentation

joerowell:
Added some documentation to enumerate_ext.h and enumerate_ext.cpp (plus some non-breaking changes to the code for cleanliness). It's probably a good idea to add some const correctness too - will try and do this when I have more time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ Windows 10 has a "Windows Subsystem for Linux", which essentially allows you to 
 
 	Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Windows-Subsystem-Linux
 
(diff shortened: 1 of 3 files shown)


cr-marcstevens:
make check-style ;)

cr-marcstevens:
Why are `tests/*.log` log files added?

joerowell:
Good point!

cr-marcstevens:
Joe: I suggest removing the last 2 commits altogether:
```
   git reset --hard HEAD~2  # remove last 2 commits locally
   git push --force         # force push to remote, removing them on github
```

And `make check-style` still fails, have you installed `clang-format`?

PR #386, opened 2019-10-31
install threadpool headers

malb:


malb:
we have

``EXTRA_DIST = io/json.hpp ballvol.const factorial.const``

I guess it might also go under

``nobase_include_fplll_HEADERS``

PR #389, opened 2019-11-01
Tutorials

marios155:
Converted test file from a .cpp file to a .sh file which compiles the three tutorials. Each one stores the result of their operations to a file, which is then cross-referenced with pre-set results (from the APIs) using the diff command. I cannot currently run make check-style as it throws this error:

clang-format -i --style=file fplll/*.{cpp,h,inl} fplll/*/*.{cpp,h,inl} tests/*.cpp
No such file or directory
Makefile:916: recipe for target 'check-style' failed
make: *** [check-style] Error 1

I also have no idea how to integrate the .sh file to the makefile, if someone could give me a tip that would be great.


tests/test_tutorials.cpp
```diff
@@ -0,0 +1,280 @@
+/*
```
malb: Should this file still be here?

tests/test_tutorials.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+cd ../tutorials
+g++ -std=c++11 -march=native -o3 lattice_base_generation_tutorial.cpp -lfplll -lmpfr -lgmp -o test_base
```
malb: I think those calls should be unified in a loop, avoiding repeating the same lines several times

tests/test_tutorials.sh
```diff
@@ -0,0 +1,24 @@
...
+g++ -std=c++11 -march=native -o3 lll_reduction_tutorial.cpp -lfplll -lmpfr -lgmp -o test__lll_reduction
+g++ -std=c++11 -march=native -o3 bkz_reduction_tutorial.cpp -lfplll -lmpfr -lgmp -o test_bkz_reduction
+mv test_base ../tests/lattices
```
malb: All of this should happen in a temporary directory using `mktemp` instead of in the source tree.

tests/test_tutorials.sh
```diff
@@ -0,0 +1,24 @@
...
+rm -f lll_output
+rm -f bkz_output
+touch base_output
+touch lll_output
+touch bkz_output
+./test_base
+./test__lll_reduction
+./test_bkz_reduction
+diff base_output base_generation_evaluation
+diff bkz_output BKZ_reduction_evaluation
```
malb: One of them should be called `_orig` the other one should perhaps be called `_gen` (or so) to make it immediately clear what is what.

malb:
Closing it, this has been inactive for a while

PR #391, opened 2019-11-04
Some basic sanity checking in MatGSO

malb:
See https://github.com/fplll/fpylll/commit/f723af4bbb05f0ca9a1ca52b643fe53218a92828

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,7 +84,7 @@ matrix:
       - sudo apt-get install -qq -y clang-format-6.0
   - os: linux
     compiler: gcc
-    env: COMPILER=g++-6 CLANGFORMAT=clang-format-6.0 CXXFLAGS="-O3 -ggdb -fprofile-arcs -ftest-coverage" 
+    env: COMPILER=g++-6 CLANGFORMAT=clang-format-6.0 CXXFLAGS="-O3 -ggdb -fprofile-arcs -ftest-coverage" CPPFLAGS="-DDEBUG"
     addons:
       apt:
         sources:
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ matrix:
       - sudo apt-get install -qq -y clang-format-3.9
   - os: linux
     compiler: gcc
-    env: COMPILER=g++-4.9 CLANGFORMAT=clang-format-6.0
+    env: COMPILER=g++-4.9 CLANGFORMAT=clang-format-6.0 CPPFLAGS="-DDEBUG"
     addons:
       apt:
         sources:
@@ -84,7 +84,7 @@ matrix:
       - sudo apt-get install -qq -y clang-format-6.0
   - os: linux
     compiler: gcc
-    env: COMPILER=g++-6 CLANGFORMAT=clang-format-6.0 CXXFLAGS="-O3 -ggdb -fprofile-arcs -ftest-coverage" CPPFLAGS="-DDEBUG"
+    env: COMPILER=g++-6 CLANGFORMAT=clang-format-6.0 CXXFLAGS="-O3 -ggdb -fprofile-arcs -ftest-coverage"
     addons:
       apt:
         sources:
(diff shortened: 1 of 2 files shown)


malb:
Don't merge yet, I'm getting many test failures when doing:

```
CPPFLAGS="-DDEBUG" ./configure --prefix=$VIRTUAL_ENV --disable-static 
make clean
make -j4 install
make check
```

```
PASS: test_nr
../test-driver: line 107:  1131 Aborted                 "$@" > $log_file 2>&1
FAIL: test_lll
../test-driver: line 107:  1145 Aborted                 "$@" > $log_file 2>&1
FAIL: test_enum
PASS: test_cvp
../test-driver: line 107:  1174 Aborted                 "$@" > $log_file 2>&1
FAIL: test_svp
../test-driver: line 107:  1187 Aborted                 "$@" > $log_file 2>&1
FAIL: test_bkz
PASS: test_pruner
PASS: test_sieve
PASS: test_gso
../test-driver: line 107:  1282 Aborted                 "$@" > $log_file 2>&1
FAIL: test_lll_gram
PASS: test_hlll
PASS: test_svp_gram
../test-driver: line 107:  1341 Aborted                 "$@" > $log_file 2>&1
FAIL: test_bkz_gram
```

Some of these are triggered by my new check, some of these seem to also exist on master.

lducas:
Shouldn't this if be a while ? (though it doesn't seem to fix the issue)

https://github.com/fplll/fplll/blob/f9436a3681f3118cd6476d88320ca11dc1b7e5f3/fplll/gso_interface.cpp#L135

kodebro:
The function `callback_set_config` in `enumerate_ext.cpp` tries to access `mu(0,0)`, which is forbidden [the diagonal of the mu-matrix is NaN]. Changing this code a bit so that mu(0,0) isn't accessed resolves already 80% of the errors. I will look at the other errors this evening.

kodebro:
TL; DR Also removing a line in the Gram GSO file fixes all errors.

Additionally removing the unnecessary  line `FPLLL_DEBUG_CHECK(i > j);` in `row_sub` in `gso_gram.cpp` resolves all errors. This function works perfectly whenever i < j as well, and during the last fplll days, @blowfish880 explained me that dual BKZ needs this functionality of this function.

Though, test_svp outputs a *lot* of text in the log-file and therefore takes quite a while to complete.

kodebro:
I made the changes in #392

malb:
I'm now only seeing my own checks fail:

```
./test_lll
test_lll: gso_interface.h:713: bool fplll::MatGSOInterface<ZT, FT>::update_gso_row(int) [with ZT = fplll::Z_NR<__mpz_struct [1]>; FT = fplll::FP_NR<__mpfr_struct [1]>]: Assertion `tmp >= 0' failed.
```

Am I checking the wrong thing?

kodebro:
To be honest, I am not completely sure. Mathematically, the r_{ii} should be positive at all times; but could it be that numerical instability spoils it? 

If I comment out your check and print the r_{ii} values, the r_{ii} sometimes are negative, but then quickly the GSO-module breaks and LLL takes over.

So, I think this isn't the preferred test to check the input matrix to be a Gram matrix. I think the best way to cope with this is to check it before-hand in high-precision, but that is computationally heavy.

malb:
Thanks, that's a bit annoying indeed. If one submits a non-Gram matrix as a Gram matrix, this might cause infinite loops and I was hoping to catch those. Seems like no dice.

kodebro:
Maybe a dumb idea, but we can try to approximate the smallest eigenvalue by means of (a small varation of) the [power iteration](https://en.wikipedia.org/wiki/Power_iteration). If this smallest eigenvalue is above zero, the matrix is positive definite.

lducas:
You can get the smallest eigenvalue e_min of a symmetric matrix A by
running principal component
analysis to get the largest eigenvalue e'_max of the symmetric A' = x * Id
- A, which is positive for a
large enough x. Then, you have e'_max = x - e_min. Taking x = 2*n^2 *
max(|A|) should suffice.

Not sure if its more numerically stable than other methods, but maybe
faster because you compute
a single eigenvalue of a symmetric matrix.

-- L

Le mar. 19 nov. 2019 à 14:41, Koen de Boer <notifications@github.com> a
écrit :

malb:
I've reverted my check for now and enabled "-DDEBUG" for one test on travis to make sure we catch bugs this way. I'll merge this PR if tests fail, seems like checking for Gram-ness is a bit more involved than we like for a cheap check

Issue #369, opened 2019-05-15
Remove the sieve?


malb:
With G6K being a thing, shall we remove the Gauss sieve again?

lgremy:
IMO, I think we need to answer (I tried to) to several questions before doing this.

1. Is the code active (i.e., commits are not just for maintenance)? It seems not, according to the log inside the sieve directory.
2. Is the code maintained? Yes
* Is it tedious to maintain the code? I do not know, but I suppose this issue is also a consequence of PR #368, in addition to the new direction sieve algorithms take with G6K
3. Is the code tested? A little
4. Is the code used as a dependency of pieces of software? I have no idea

If I correctly understood (I am not a specialist), with [the G6K implementation](https://github.com/fplll/g6k), G6K outperforms the known results about practical sieving, and then the sieve implementation inside fplll introduced with [BLS16]. However, since this code is linked to a research article, if we remove the Gauss sieve implementation from fplll, I suggest that we will freeze the code in a specific and well-identified branch, in order for a further use to be an option.

shi-bai:
Hi Laurent, Martin,

Thanks for bringing this up.
On Fri, May 17, 2019 at 3:10 AM Laurent Grémy <notifications@github.com>
wrote:

> IMO, I think we need to answer (I tried to) to several questions before
> doing this.
>
>    1. Is the code active (i.e., commits are not just for maintenance)? It
>    seems not, according to the log inside the sieve directory.
>    2. Is the code maintained? Yes
>
>
>    - Is it tedious to maintain the code? I do not know, but I suppose
>    this issue is also a consequence of PR #368
>    <https://github.com/fplll/fplll/pull/368>, in addition to the new
>    direction sieve algorithms take with G6K
>
>
>    1. Is the code tested? A little
>    2. Is the code used as a dependency of pieces of software? I have no
>    idea
>
> If I correctly understood (I am not a specialist), with the G6K
> implementation <https://github.com/fplll/g6k>, G6K outperforms the known
> results about practical sieving, and then the sieve implementation inside
> fplll introduced with [BLS16]. However, since this code is linked to a
> research article, if we remove the Gauss sieve implementation from fplll, I
> suggest that we will freeze the code in a specific and well-identified
> branch, in order for a further use to be an option.
>
This looks good to me. I think we could leave the code in a specific branch
(just in case anyone wanted to read the code with the paper).

Cheers,
Shi

malb:
[...]

PR #393, opened 2019-11-19
Optional latsieve isn't a thing anymore.

joerowell:
When running the "make latsieve command" I get a "no rule" error - it appears that this isn't a valid target anymore, and that we always build the sieve directory & related classes under the regular make command.
 
As a result, I've removed the "make latsieve" sentence.

This might be irrelevant though if issue #369 is finalised before the next release.


joerowell:
Wow I was so wrong - no idea what happened there!

joerowell:
This also now adds two new methods to NumVect - a constructor that takes a std::vector input, and an overload for ==.

PR #398, opened 2019-11-29
Fix out-of-bounds vector accesses in Pruner::enforce.

jamesjer:
I have been testing builds for fplll 5.3.0 and fpylll 0.5.0dev for Fedora.  The default Fedora build flags include -D_GLIBCXX_ASSERTIONS.  This led to test failures in fpylll due to out of bounds vector accesses in the fplll code.  This commit fixes the two such out of bounds accesses found by the fpylll tests.  (I highly recommend that you build with -D_GLIBCXX_ASSERTIONS before running your tests, by the way.)

In the first hunk, note that min_pruning_coefficients always has length *d*, not n.  The fpylll test suite, at least, invokes this code in such a way that i / c can be equal to d and therefore one past the end of the min_pruning_coefficients vector.

In the second hunk, j can be as large as as dn (b.size()), and therefore on the first iteration of the loop, when i is j - 1, then b[i + 1] == b[j] is one past the end of b.

### DIFF
--- a/fplll/pruner/pruner.h
+++ b/fplll/pruner/pruner.h
@@ -1027,8 +1027,8 @@ template <class FT> inline bool Pruner<FT>::enforce(/*io*/ vec &b, /*opt i*/ con
     status |= (b[i] > 1.0001);
     b[i] = b[i] > 1 ? 1. : b[i];
 
-    // note min_pruning_coefficients always has length n
-    if (b[i] <= min_pruning_coefficients[i / c])
+    // note min_pruning_coefficients always has length d
+    if (i / c < d && b[i] <= min_pruning_coefficients[i / c])
       b[i] = min_pruning_coefficients[i / c];
   }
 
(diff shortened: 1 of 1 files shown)


shi-bai:
James, thanks for the report. 

On my debian, I cannot trigger the error by enabling D_GLIBCXX_ASSERTIONS in configuration. The tests seem to pass.

============================================================================
Testsuite summary for fplll 5.3.0
============================================================================
# TOTAL: 13
# PASS:  13
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0

Does the checking error comes from Line 22 of test_pruner.cpp?

Shi

malb:
Over at https://github.com/fplll/fpylll/issues/161 it seems to remove some non deterministic behaviour, so I'm inclined to merge this soon. Shi, do you sign off on this?

shi-bai:
I could not repeat this error on my debian. But see fplll/fpylll#161 that removes error. Happy to merge.

malb:
Ta!

PR #407, opened 2020-02-12
configure.ac: improved libqd detection

orlitzky:
Here's my first shot at this. I'm sure it will look buggy as hell in the morning. Details in the individual commit messages.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -115,10 +115,10 @@ AS_IF([test "x$with_qd" != "xno"], [
       ])
 
 AS_IF([test "x${have_libqd}" = "xyes"], [
-    LIBQD_LIBADD="-lqd"
+    LIBQD_LIBS="-lqd"
     AC_DEFINE([FPLLL_WITH_QD], [1], [defined when libqd-dev is available])])
 
-AC_SUBST(LIBQD_LIBADD)
+AC_SUBST(LIBQD_LIBS)
 
 # Checks for header files.
 AC_HEADER_STDC
--- a/configure.ac
+++ b/configure.ac
@@ -106,17 +106,19 @@ AC_CHECK_LIB(mpfr, mpfr_fms, , [AC_MSG_ERROR(
 AC_ARG_WITH(qd, AS_HELP_STRING([--with-qd=@<:@=DIR@:>@], [quaddouble install directory]),)
 
 AS_IF([test "x$with_qd" != "xno"], [
-       CPPFLAGS="$CPPFLAGS -I$with_qd/include"
-       LDFLAGS="$LDFLAGS -L$with_qd/lib"
-       qd_lib_path="$with_qd/lib"
-       AC_CHECK_HEADER(qd/dd_real.h,
-           have_libqd="yes",
-           have_libqd="no")
-      ])
+  CPPFLAGS="$CPPFLAGS -I$with_qd/include"
+  AC_CHECK_HEADER(qd/dd_real.h, [
+    LDFLAGS="$LDFLAGS -L$with_qd/lib"
+    qd_lib_path="$with_qd/lib"
+    LIBQD_LIBS="-lqd"
+    have_libqd="yes"],
+    [ AC_MSG_ERROR([unable to find header qd/dd_real.h]) ]
+  )]
+)
 
 AS_IF([test "x${have_libqd}" = "xyes"], [
-    LIBQD_LIBS="-lqd"
-    AC_DEFINE([FPLLL_WITH_QD], [1], [defined when libqd-dev is available])])
+  AC_DEFINE([FPLLL_WITH_QD], [1], [defined when libqd is usable])
+])
 
 AC_SUBST(LIBQD_LIBS)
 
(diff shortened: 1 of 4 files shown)


malb:
Looks good to me, did you find any bugs on second reading? If not, I'd merge and deal with the fallout if it arises.

orlitzky:
A few things are still a little weird, but it's all in the code I didn't touch. I'll be around to fix it if something doesn't work as expected. I think I've checked all possible cases:

* qd installed
  * `./configure`
  * `./configure --without-qd`
  *  `./configure --with-qd`
  *  `./configure --with-qd=/path/to/qd`
* qd installed, but with the pkgconfig (*.pc) file deleted
  * `./configure`
  * `./configure --without-qd`
  *  `./configure --with-qd`
  *  `./configure --with-qd=/path/to/qd`
* qd NOT installed
  * `./configure`
  * `./configure --without-qd`
  *  `./configure --with-qd`
  *  `./configure --with-qd=/path/to/qd`

I pushed one additional commit with only whitespace changes.

malb:
Can we make `PKG_CHECK_MODULES` optional? Or maybe we should just require it, it should be fairly standard by now. Thoughts?

orlitzky:
> Can we make `PKG_CHECK_MODULES` optional? Or maybe we should just require it, it should be fairly standard by now. Thoughts?

When you run `make dist`, it should create a tarball with an `aclocal.m4` that contains the contents of `pkg.m4`, namely the `PKG_CHECK_MODULES` macro. So long as that's true, pkg-config is really only non-optional for developers, which I think is acceptable.

malb:
Thanks, that's a good point.

malb:
Hi, two more issues:
- It seems we now don't build with `qd` support by default, i.e. when `qd` is detected but `--with-qd` is not given. Building with `qd` should be the default
- We also don't seem to add `-lqd` to FPLLL's pkg-config file, which means `fpylll` doesn't pick up `qd`.

orlitzky:
> We also don't seem to add `-lqd` to FPLLL's pkg-config file, which means `fpylll` doesn't pick up `qd`.

This should work if libqd is being used, it's in https://github.com/fplll/fplll/blob/master/fplll.pc.in#L10
I checked it with both `--with-qd` and `--with-qd=/usr` on my machine.

> It seems we now don't build with `qd` support by default, i.e. when `qd` is detected but `--with-qd` is not given. Building with `qd` should be the default

I made that choice but it's easy to change. My reasoning was that if the autodetection is enabled by default, you can run the same `./configure` command on two different machines and wind up with two very different libraries. Defaulting to "yes" or "no" is less stochastic, and of those only "no" is a workable option.

In any case, the default is set in https://github.com/fplll/fplll/blob/master/configure.ac#L109 and you should be able to change it to "yes" without hurting anything.

malb:
Adding `-lqd` doesn't seem to work for me:

```
configure:13252: checking for LIBQD
configure:13259: $PKG_CONFIG --exists --print-errors "qd"
Package qd was not found in the pkg-config search path.
Perhaps you should add the directory containing `qd.pc'
to the PKG_CONFIG_PATH environment variable
No package 'qd' found
configure:13262: $? = 1
configure:13276: $PKG_CONFIG --exists --print-errors "qd"
Package qd was not found in the pkg-config search path.
Perhaps you should add the directory containing `qd.pc'
to the PKG_CONFIG_PATH environment variable
No package 'qd' found
configure:13279: $? = 1
configure:13293: result: no
No package 'qd' found
configure:13311: checking for library containing c_dd_add
configure:13342: g++ -std=c++11 -o conftest -O3 -Wall   conftest.cpp -lmpfr -lmpfr -lgmp -lgmp  >&5
/usr/bin/ld: /tmp/ccfaD8FB.o: in function `main':
conftest.cpp:(.text.startup+0x5): undefined reference to `c_dd_add'
collect2: error: ld returned 1 exit status
configure:13342: $? = 1
configure: failed program was:
<…>
configure:13342: g++ -std=c++11 -o conftest -O3 -Wall   conftest.cpp -lqd  -lmpfr -lmpfr -lgmp -lgmp  >&5
configure:13342: $? = 0
configure:13359: result: -lqd
configure:13365: checking qd/dd_real.h usability
configure:13365: g++ -std=c++11 -c -O3 -Wall  conftest.cpp >&5
configure:13365: $? = 0
configure:13365: result: yes
configure:13365: checking qd/dd_real.h presence
configure:13365: g++ -E -std=c++11  conftest.cpp
configure:13365: $? = 0
configure:13365: result: yes
configure:13365: checking for qd/dd_real.h
configure:13365: result: yes
configure:13492: checking for ANSI C header files
configure:13596: result: yes
```

I then have:

```
$ grep QD fplll/fplll_config.h
#define FPLLL_WITH_QD 1
$  grep Libs fplll.pc
Libs: -L${libdir}  -lgmp -lpthread -lmpfr -lfplll
```

orlitzky:
> Adding -lqd doesn't seem to work for me

Ok, I figured out the problem. I'll fix that and try to clean up the other repeated `-lfoo` values in the pkg-config file as well.

PR #411, opened 2020-03-06
Fix the libraries listed in the pkg-config file

orlitzky:
When no pkg-config file is present and when qd was automatically detected, the `-lqd` flag was winding up in `LIBS` rather than `LIBQD_LIBS`. The other linker flags that arise from the `AC_CHECK_LIB` macro also wind up in `LIBS`, so it makes sense to put `LIBS` in the pkg-config file, replacing some of the existing "custom" entries. The remaining commits attempt to eliminate the duplicate flags. Hopefully the individual commit messages explain.


malb:
Thanks!

PR #419, opened 2020-04-14
check for -latomic

### LINKED ISSUE #418: add `-latomic` when needed
The compilation on an armhf machine (RPi4) fails with
```
/usr/bin/ld: ./.libs/libfplll.so: undefined reference to `__atomic_store_8'                                                                     /usr/bin/ld: ./.libs/libfplll.so: undefined reference to `__atomic_load_8'
```
Apparently it needs to be linked with `libatomic`. Adding `LDFLAGS="$LDFLAGS -latomic"` to `configure.ac` solves the issue for me. Sagemath has recently updated the version of fplll it uses and as of it is broken on armhf too.
gcc 8.3.0

malb:
fixes #418

----

#

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -155,6 +155,8 @@ AC_CHECK_FUNCS([floor pow rint sqrt strtol])
 
 AX_PTHREAD()
 
+CHECK_ATOMIC()
+
 # enumeration dimension
 max_enum_dim=256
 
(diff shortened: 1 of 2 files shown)


cr-marcstevens:
Looks good to me. I'm fine with merging when confirmed by Travis and solves issue #418 by @IvaKom

IvaKom:
This modification solves the problem for me, thank you!

Issue #417, opened 2020-04-13
A new  Z_NR<mpz>constructor?


johughes99:
I want to create an object of type Z_NR<mpz_t> and then assign it a value from a mpz_t object.

I had hoped that there was a Z_NR<mpz_t> constructor to do something like:

`	mpz_t tmp_mpz;
	mpz_init (tmp_mpz);
	// assign zValue with the value of tmp_mpz
	Z_NR<mpz_t> zValue = Z_NR<mpz_t>(tmp_mpz);`

However, this does not seem to be possible. 

Of course, something like:

`	Z_NR<mpz_t> zOrder = Z_NR<mpz_t>();
	zOrder = tmp_mpz;`

works fine. But having a constructor to do this would be nicer.

malb:
I agree this would be good to have. PRs welcome :)

johughes99:
Maybe - when I get more familar with the code.

Also, it would be nice to have a modulus operator on integers of type Z_NR - as I couldn't find one.

joerowell:
@malb This can be closed, right? It's now in master as of #422.

PR #421, opened 2020-04-21
Added constructors from  #417

joerowell:
#417 asks for copy-assignment/move-assignment constructors for the integer types.
Thankfully, C++ compilers mostly implement this for us; the only important bit is adding in the copy constructors (I didn't add in move-assignment because I don't think it's used anywhere else in fplll...).

This PR implements #417 . It's now possible to write:

``` 
mpz_t temp;
mpz_init(temp, 5);
Z_NR<mpz_t> out(temp)
```
and:

``` Z_NR<mpz_t> out = Z_NR<mpz_t>(temp); ```

as requested. 

It also does it for all of the integral types - so mpz_t above could be replaced by ```long``` or ```double``` and it would still work.

Note; I didn't implement the request for a % operator.

### DIFF
--- a/fplll/nr/nr_Z_mpz.inl
+++ b/fplll/nr/nr_Z_mpz.inl
@@ -11,7 +11,7 @@ FPLLL_BEGIN_NAMESPACE
 template <> inline Z_NR<mpz_t>::Z_NR() { mpz_init(data); }
 
 template <> inline Z_NR<mpz_t>::Z_NR(const Z_NR<mpz_t> &z) { mpz_init_set(data, z.data); }
-
+template <> inline Z_NR<mpz_t>::Z_NR(const mpz_t &z) {mpz_init_set(data,z);}
 template <> inline Z_NR<mpz_t>::~Z_NR() { mpz_clear(data); }
 
 /** get data */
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: fplll/nr/nr_Z_mpz.inl [hunk 1]
```diff
@@ -11,7 +11,7 @@ FPLLL_BEGIN_NAMESPACE
 template <> inline Z_NR<mpz_t>::Z_NR() { mpz_init(data); }
 
 template <> inline Z_NR<mpz_t>::Z_NR(const Z_NR<mpz_t> &z) { mpz_init_set(data, z.data); }
-
+template <> inline Z_NR<mpz_t>::Z_NR(const mpz_t &z) { mpz_init_set(data, z); }
```
malb, line 14: I think this needs documentation that this does not take ownership of z, i.e.the caller will have to still `mpz_free` it after.

joerowell: Great catch - I've added it. Adding move constructors in future would be useful, but doing this (and testing it doesn't implicitly break anything) is probably an Fplll days task...

malb: You added it to `operator=` but shouldn't we add it to both?

joerowell: I've agree - it's been added.

PR #422, opened 2020-04-24
Added modulus operation 

joerowell:
This is the second part from #417 - Z_NR now has a modulus operation. 

The implementation merely adds wrappers for Z_NR; internally, each method delegates to the modulus operations for the underlying types.

Usage is very similar to all of the other operations (add, mul etc). 
For example: 
```
Z_NR<mpz_t> out = 0;
Z_NR<mpz_t> radix = 5;
Z_NR<mpz_t> modulus = 3;
out.mod(radix, modulus);
```
Note that the template type parameters must be the same; I think mixing the types is a bad idea.

### DIFF
--- a/fplll/nr/nr_Z.inl
+++ b/fplll/nr/nr_Z.inl
@@ -157,6 +157,11 @@ public:
   inline void sub_ui(const Z_NR<Z> &a, unsigned int b);
 
   /**
+   * value := a mod b
+   */
+  inline void mod(const Z_NR<Z> &a, const Z_NR<Z> &b);
+
+  /**
    * value := -a.
    */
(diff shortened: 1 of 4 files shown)


malb:
Shouldn't it act on `value`? It seems a bit strange to not read `value` but `write` it.

joerowell:
So, what you're proposing is to simplify it to (say) a.mod(b)? Am I
understanding your point correctly?

The reason I didn't do it the way you've suggested is for consistency with
the other arithmetic operations in Z_NR (I.e add, mul etc all follow this
signature).

Happy to change it if you'd prefer it this way though; I agree it's a more
natural usage.

malb:
Good point.

PR #433, opened 2020-06-25
Add continuous integration tests with SageMath (multi-platform) 

mkoeppe:
See https://trac.sagemath.org/ticket/29091

Preview of a successful run on `raspbian-buster-armhf-standard` at ​https://github.com/mkoeppe/fplll/runs/805921434

### DIFF
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -30,8 +30,10 @@ env:
   TARGETS_PRE: build/make/Makefile
   TARGETS:     SAGE_CHECK=yes fplll
   TARGETS_OPTIONAL: build/make/Makefile
-  SAGE_REPO:   sagemath/sage
-  SAGE_REF:    develop
+  SAGE_REPO:   mkoeppe/sage
+  SAGE_REF:    ci-29091
+  # SAGE_REPO:   sagemath/sage
+  # SAGE_REF:    develop
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


malb:
I think a five hour wait time for CI is asking a bit too much. Also, this should be for FPyLLL not FPLLL?

mkoeppe:
You can configure it to run only on pull requests and tags, rather than on every commit. See commented out stuff at the beginning of the file.

mkoeppe:
Here's a better version. Does not use a private repository any more (gets a branch from trac.sagemath.org instead), runs on pull requests and tags only, and adds macOS testing (homebrew + conda)

malb:
We've switched to GitHub Actions now. I'd be up for merging this triggered manually only, i.e. on `workflow_dispatch`.

mkoeppe:
> We've switched to GitHub Actions now. I'd be up for merging this triggered manually only, i.e. on `workflow_dispatch`.

Sounds good. Would you like to update the pull request?

malb:
That would be great!

malb:
Oh, another thing: shouldn't this be against FPyLLL? Typically, we release new versions of FPLLL and FPyLLL together and the latter depends on the former.

.github/workflows/ci-sage.yml
```diff
@@ -0,0 +1,337 @@
...
+  # Name of this project in the Sage distribution
+  SPKG:        fplll
+  # Sage distribution packages to build
+  TARGETS_PRE: build/make/Makefile
+  TARGETS:     SAGE_CHECK=yes fplll
+  TARGETS_OPTIONAL: build/make/Makefile
+  # Standard setting: Test the current beta release of Sage:
+  SAGE_REPO:   sagemath/sage
+  SAGE_REF:    develop
+  # Temporarily test on the branch from sage ticket 29091
```
malb, line 58: How would this be handled in general?

.github/workflows/ci-sage.yml
```diff
@@ -0,0 +1,337 @@
...
+        if: always()
+      - uses: actions/upload-artifact@v1
+        with:
+          path: artifacts
+          name: ${{ env.LOGS_ARTIFACT_NAME }}
+        if: always()
+      - name: Print out logs for immediate inspection
+        # and markup the output with GitHub Actions logging commands
+        run: |
+          .github/workflows/scan-logs.sh "artifacts/$LOGS_ARTIFACT_NAME"
```
malb, line 336: Is this file missing?

mkoeppe:
> Oh, another thing: shouldn't this be against FPyLLL? Typically, we release new versions of FPLLL and FPyLLL together and the latter depends on the former.

OK, I'll send you a pull request against fpylll and update this one. You can then decide whether you want both or only one of them.

.github/workflows/ci-sage.yml
```diff
@@ -0,0 +1,337 @@
...
+    types: [opened, synchronize]
+  push:
+    tags:
+      - '*'
+
+env:
+  # Ubuntu packages to install so that the project's "make dist" can succeed
+  DIST_PREREQ: libgmp-dev autoconf automake libtool libmpfr-dev
+  # Name of this project in the Sage distribution
+  SPKG:        fplll
```
malb, line 50: For completeness, this should create new SPKGs for both `fplll` and `fpylll` from the latest tags as those are typically updated together.

Issue #435, opened 2020-07-01
avoid cputime() calls


malb:
Paul Zimmermann reports:

> In short, you have to change line 52 of fplll/lll.h (in fplll-5.3.2) as follows:
>
> -  int start_time = cputime();
> +  int start_time = (verbose) ? cputime() : 0;
>
> The reason is that cputime() calls getrusage(), and calling getrusage() in a multi-thread environment is a disaster. I guess you can do the same in the other functions that call cputime.

That is, calls to `cputime()` have a negative effect on multi-threaded performance. 

I see these occurrences:

``` shell
fplll/bkz.cpp:33:      cputime_start(0)
fplll/bkz.cpp:376:             (cputime() - cputime_start) * 0.001);
fplll/bkz.cpp:459:             (cputime() - cputime_start) * 0.001);
fplll/bkz.cpp:511:             (cputime() - cputime_start) * 0.001);
fplll/bkz.cpp:562:  cputime_start = cputime();
fplll/bkz.cpp:587:    if ((flags & BKZ_MAX_TIME) && (cputime() - cputime_start) * 0.001 >= param.max_time)
fplll/bkz.cpp:668:    dump_gso(param.dump_gso_filename, true, "Output", -1, (cputime() - cputime_start) * 0.001);
fplll/bkz.cpp:683:       << std::setw(9) << std::setprecision(3) << (cputime() - cputime_start) * 0.001 << "s";
fplll/lll.cpp:52:  int start_time = cputime();
fplll/lll.cpp:89:             << " cputime=" << cputime() - start_time << endl;
fplll/hlll.cpp:36:  int start_time = cputime();
fplll/hlll.cpp:45:    cerr << "Discovering vector 1/" << m.get_d() << " cputime=" << cputime() - start_time << endl;
fplll/hlll.cpp:71:    cerr << "Discovering vector 2/" << m.get_d() << " cputime=" << cputime() - start_time << endl;
fplll/hlll.cpp:125:                 << " cputime=" << cputime() - start_time << endl;
```

joerowell:
Sorry if it's a stupid question: why do we care about the CPU time in these functions? That's probably the best way to decide upon a viable solution (other than just using the ``verbose`` flag).

malb:
We use it for statistics and in BKZ for early abort, i.e. you can say: "run it for X time" (not sure anybody is using this)

joerowell:
Fixed in #438 .

PR #438, opened 2020-07-16
Add verbose guard to cputime in hlll, lll and bkz

joerowell:
This pull request addresses #435 .

## What changes?
* Add ```verbose``` guard to calls to ```cputime``` in ```hlll.cpp```, ```lll.cpp``` and ```bkz.cpp```. Some of the calls mentioned in #435 are already wrapped inside an ```if``` statement that checks ```verbose```, so it doesn't make sense to double wrap these.

## What doesn't it change?
There's no new flag. In particular, in ```bkz.cpp``` every call relies exclusively on ```BKZ_DUMP_GSO```.

The reason why is because every  ```cputime```  call (other than the one changed in this PR) is already inside a check on ```BKZ_DUMP_GSO```; I don't think it makes sense to remove the call to ```cputime``` there, because it's probably a sensible default for ```BKZ_DUMP_GSO```. This is compounded by the fact that the call to ```print_tour``` is inside a check on ```BKZ_DUMP_GSO```; so this seems like expected behaviour. 

If it's necessary, I'm happy to add another parameter (something like ```BKZ_DUMP_GSO_NO_TIME```), or to change the default behaviour of ```BKZ_DUMP_GSO``` but this is something I wanted to check first (as it's a change in the current behaviour).

### DIFF
--- a/fplll/bkz.cpp
+++ b/fplll/bkz.cpp
@@ -559,7 +559,8 @@ template <class ZT, class FT> bool BKZReduction<ZT, FT>::bkz()
     print_params(param, cerr);
     cerr << endl;
   }
-  cputime_start = cputime();
+  
+  cputime_start = (flags & BKZ_DUMP_GSO) ? cputime() : 0;
 
   m.discover_all_rows();
 
--- a/fplll/bkz.cpp
+++ b/fplll/bkz.cpp
@@ -559,8 +559,8 @@ template <class ZT, class FT> bool BKZReduction<ZT, FT>::bkz()
     print_params(param, cerr);
     cerr << endl;
   }
-  
-  cputime_start = (flags & BKZ_DUMP_GSO) ? cputime() : 0;
+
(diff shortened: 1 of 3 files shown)


fplll/bkz.cpp
```diff
@@ -559,7 +559,8 @@ template <class ZT, class FT> bool BKZReduction<ZT, FT>::bkz()
     print_params(param, cerr);
     cerr << endl;
   }
-  cputime_start = cputime();
+
+  cputime_start = (flags & BKZ_DUMP_GSO) ? cputime() : 0;
```
malb, line 563: I think this should also use `BKZ_MAX_TIME`:

```c++
if ((flags & BKZ_MAX_TIME) && (cputime() - cputime_start) * 0.001 >= param.max_time)
```

joerowell: Good catch: I've added ``` flags & BKZ_MAX_TIME``` to the ternary.

PR #440, opened 2020-08-08
Add customisation on the node counting

joerowell:
## What does this pull request do?
At the moment, fplll counts every node during enumeration into a single ``uint64_t`` counter. This approach is a good default, but there may be situations where other counting mechanisms are necessary (i.e counting every node on a given level).
This pull request attempts to offer this behaviour whilst (a) providing an extensible interface for future adaptation, (b) not compromising much on runtime performance and (c) not breaking existing code too much.

## How does this pull request work?
The approach I've gone with is the policy pattern. Briefly, the idea is similar to dynamic dispatch: each Enumeration class gets an abstraction of a counter, ```CounterClassWrapper```, and all calls to update the counter go through this interface.
However, because the counter class is constantly updated, it may be expensive to use the traditional approach of virtual functions. To get around this, I've used templates to provide this abstraction: this should enable the compiler to be clever and generate object code that would, in theory, be close to what one might write by hand.

Let me give you an example. The traditional approach using virtual functions might look like this:
```
class Enumerator {
              void Enumerate(int dim, .... , Counter& counter); 
}; 
```

Instead of passing the Counter at runtime, I've implemented the following:
```
template<typename Counter>
class Enumerator {
             void Enumerate(int dim,...);
             CounterClassWrapper<Counter> _node_counter; 
};
```

Note: the above example uses a wrapper for the Counter. In practice, this delegates to the methods defined in the Counter. This is to provide type safety: there's no restrictions on the class that may be passed as a Counter, but doing it this way ensures that certain methods need to be implemented. 

In particular, this pull request has 3 new classes:
* ```WholeTreeCounter```. This counter is the default behaviour: there's a single ```uint64_t``` that counts the whole tree. At the moment this is the project wide default.
* ```LevelTreeCounter```.  This counter counts the number of nodes visited at each level, storing them in an array.
* ```CounterClassWrapper```. This class is a wrapper for the counters, as described above. 

## What's left to do?
This pull request isn't finished. This is because the parallel enumeration code hasn't been adapted yet for this usage.
In theory, this would be easy to do: the parallel enumeration code already counts the number of entries per level, and so it would be easy to transfer it over to both of the counters I've implemented.

The reason why I haven't is because there's two different size definitions that are used: ```FPLLL_MAX_ENUM_DIM``` for the whole of fplll, and ```FPLLL_MAX_PARALLEL_ENUM_DIM``` for the parallel enumerator. 

I wanted to ask: do we expect that ```FPLLL_MAX_PARALLEL_ENUM_DIM``` will always be smaller than ```FPLLL_MAX_ENUM_DIM```? If this is true then the fix shouldn't take long at all.


cr-marcstevens:
Regarding your last question: yes, the max parallel enum dim in theory should always be smaller/equal to max enum dim.
This is because there is a heavy compile cost related to the former, so there one wants to be more conservative.
However, at the moment it is not guaranteed, so a static_assert can be used to force this behaviour.

joerowell:
> Regarding your last question: yes, the max parallel enum dim in theory should always be smaller/equal to max enum dim.
> This is because there is a heavy compile cost related to the former, so there one wants to be more conservative.
> However, at the moment it is not guaranteed, so a static_assert can be used to force this behaviour.

OK: I'll add this when I get time (probably Monday).

joerowell:
Whilst Travis has passed, I'm going to write a separate test file for this system. I'll update the PR when it's done.

malb:
Thanks Joe. We should also wait with merging, even after a positive review, until I had a chance to update FPyLLL to match this just to catch potential issues from that.

malb:
It seems this PR also significantly increases compilation time (or is this just subjective). If so, I'm not sure we want to pay this price for backward compatibility, i.e. we can break the API, this is a research library.

joerowell:
It's likely that's true: we do instanstiate a lot of extra templates.

If the compile time is an issue, then I'm happy to pull this pr & come back
with an approach that's similar to the evaluators. It shouldn't be *that*
much slower. Failing that, we could just switch over to only returning
arrays and break the API?

malb:
I vote for breaking the API. Code simplicity is much more valuable for this sort of software.

malb:
I tested it on holzfusion single core:

- master:

```
make  542.98s user 11.53s system 99% cpu 9:14.64 total
```

- this pr:

```
make  635.69s user 13.62s system 99% cpu 10:49.43 total
```

so whether this is too much is debatable. 

Still, in terms of API/simplicity. Wouldn't it be "easier" to always record as nodes per level and and change the signature of `get_nodes()` to `get_nodes(level=-1)`. It would return an `uint64_t` which is the nodes per level for `level >= 0` and the sum of them otherwise.

joerowell:
It would be, but there's no extensibility in that. For example, if you later wanted to count the nodes differently then it would require an entire API change. However, I'm not opposed to it, so I'll modify this PR with your changes in mind.

malb:
Hi Joe, first, sorry, I should have thought about this more carefully much earlier. I think the lack of flexibility is acceptable here. It's hard to image how else we'd want to count them?

PR #442, opened 2020-08-15
Count nodes by level, rather than by tree

joerowell:
Following #440, this PR changes the counters in fplll from a single, ```uint64_t``` to an array, which contains a counter for each level. 

In particular:
* ```get_nodes()``` has become ```get_nodes(const int level=-1)```. If ```level=-1``` then the sum of all counters is returned.
* ```get_nodes_array()``` now exists. This primarily exists to return the counters from the external enumerator.

### DIFF
--- a/fplll/enum/enumerate.h
+++ b/fplll/enum/enumerate.h
@@ -117,6 +117,8 @@ public:
     return _nodes[level];
   }
 
+  inline array<uint64_t, EnumerationBase::maxdim> get_nodes_array() const { return _nodes; }
+
 private:
   MatGSOInterface<ZT, FT> &_gso;
   Evaluator<FT> &_evaluator;
--- a/fplll/enum/enumerate.h
+++ b/fplll/enum/enumerate.h
@@ -36,6 +36,7 @@ public:
       : _gso(gso), _evaluator(evaluator)
   {
     _max_indices = max_indices;
+    std::fill(nodes.begin(), nodes.end(), 0);
   }
 
   void enumerate(int first, int last, FT &fmaxdist, long fmaxdistexpo,
@@ -44,7 +45,16 @@ public:
                  const vector<enumf> &pruning = vector<enumf>(), bool dual = false,
                  bool subtree_reset = false);
 
-  inline uint64_t get_nodes() const { return nodes; }
+  inline uint64_t get_nodes(const int level = -1) const
+  {
+    if (level == -1)
+    {
+      return std::accumulate(nodes.cbegin(), nodes.cend(), 0);
+    }
+    return nodes[level];
+  }
+
+  inline array<uint64_t, FPLLL_MAX_ENUM_DIM> get_nodes_array() { return nodes; }
 
 private:
   MatGSOInterface<ZT, FT> &_gso;
@@ -70,7 +80,7 @@ template <typename ZT, typename FT> class Enumeration
 public:
   Enumeration(MatGSOInterface<ZT, FT> &gso, Evaluator<FT> &evaluator,
               const vector<int> &max_indices = vector<int>())
-      : _gso(gso), _evaluator(evaluator), _max_indices(max_indices), enumdyn(nullptr), _nodes(0)
+      : _gso(gso), _evaluator(evaluator), _max_indices(max_indices), enumdyn(nullptr), _nodes{}
   {
   }
 
@@ -87,7 +97,7 @@ public:
         enumext.reset(new ExternalEnumeration<ZT, FT>(_gso, _evaluator));
       if (enumext->enumerate(first, last, fmaxdist, fmaxdistexpo, pruning, dual))
       {
-        _nodes = enumext->get_nodes();
+        _nodes = enumext->get_nodes_array();
         return;
       }
     }
@@ -97,10 +107,15 @@ public:
       enumdyn.reset(new EnumerationDyn<ZT, FT>(_gso, _evaluator, _max_indices));
     enumdyn->enumerate(first, last, fmaxdist, fmaxdistexpo, target_coord, subtree, pruning, dual,
                        subtree_reset);
-    _nodes = enumdyn->get_nodes();
+    _nodes = enumdyn->get_nodes_array();
   }
 
-  inline uint64_t get_nodes() const { return _nodes; }
+  inline uint64_t get_nodes(const int level = -1) const
+  {
+    if (level == -1)
+      return std::accumulate(_nodes.begin(), _nodes.end(), 0);
+    return _nodes[level];
+  }
 
 private:
   MatGSOInterface<ZT, FT> &_gso;
@@ -108,7 +123,7 @@ private:
   vector<int> _max_indices;
   std::unique_ptr<EnumerationDyn<ZT, FT>> enumdyn;
   std::unique_ptr<ExternalEnumeration<ZT, FT>> enumext;
-  uint64_t _nodes;
+  array<uint64_t, EnumerationBase::maxdim> _nodes;
 };
 
 FPLLL_END_NAMESPACE
--- a/fplll/enum/enumerate_base.cpp
+++ b/fplll/enum/enumerate_base.cpp
@@ -30,7 +30,7 @@ inline void EnumerationBase::enumerate_recursive(
 
   if (!(newdist <= partdistbounds[kk]))
     return;
-  ++nodes;
+  ++nodes[kk];
 
   alpha[kk] = alphak;
   if (findsubsols && newdist < subsoldists[kk] && newdist != 0.0)
@@ -87,7 +87,7 @@ inline void EnumerationBase::enumerate_recursive(
       enumf newdist2 = partdist[kk] + alphak2 * alphak2 * rdiag[kk];
       if (!(newdist2 <= partdistbounds[kk]))
         return;
-      ++nodes;
+      ++nodes[kk];
       alpha[kk] = alphak2;
       if (kk == 0)
       {
@@ -118,7 +118,7 @@ inline void EnumerationBase::enumerate_recursive(
       enumf newdist2 = partdist[kk] + alphak2 * alphak2 * rdiag[kk];
       if (!(newdist2 <= partdistbounds[kk]))
         return;
-      ++nodes;
+      ++nodes[kk];
       alpha[kk] = alphak2;
       if (kk == 0)
       {
@@ -190,7 +190,7 @@ template <bool dualenum, bool findsubsols, bool enable_reset> void EnumerationBa
 
   partdist[k_end] = 0.0;  // needed to make next_pos_up() work properly
 
-  nodes -= k_end - k;
+  nodes[k_end] -= k_end - k;
   k = k_end - 1;
 
 #ifdef FPLLL_WITH_RECURSIVE_ENUM
@@ -207,7 +207,7 @@ template <bool dualenum, bool findsubsols, bool enable_reset> void EnumerationBa
                            << " newdist=" << newdist << " partdistbounds_k=" << partdistbounds[k]);
     if (newdist <= partdistbounds[k])
     {
-      ++nodes;
+      ++nodes[k];
       alpha[k] = alphak;
       if (findsubsols && newdist < subsoldists[k] && newdist != 0.0)
       {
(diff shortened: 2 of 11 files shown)


fplll/enum/enumerate.h
```diff
@@ -44,7 +45,16 @@ template <typename ZT, typename FT> class EnumerationDyn : public EnumerationBas
...
                  bool subtree_reset = false);
 
-  inline uint64_t get_nodes() const { return nodes; }
+  inline uint64_t get_nodes(const int level = -1) const
+  {
+    if (level == -1)
+    {
+      return std::accumulate(nodes.cbegin(), nodes.cend(), 0);
+    }
+    return nodes[level];
```
malb, line 54: Do we just want to accept this buffer overread or shall we do a bounds check?

malb:
Did you run tests on this? I'm getting:

```python
import math
from fpylll import *
FPLLL.set_random_seed(1337)
A = LLL.reduction(IntegerMatrix.random(40, "qary", k=20, bits=30))
M = GSO.Mat(A)
M.update_gso()
pr = Pruning.Pruner(0.99*M.get_r(0,0), 2**20, [M.r()], 0.51)
c = pr.optimize_coefficients([1. for _ in range(M.d)])
cost, details = pr.single_enum_cost(c, True)
lft = [int(math.ceil(x)) for x in details]
enum_obj = Enumeration(M)
v = enum_obj.enumerate(0, M.d, 0.99*M.get_r(0,0), 0)
rgt = [enum_obj.get_nodes(i) for i in range(M.d)]

print(sum(lft), int(math.ceil(cost)))
print(sum(rgt), enum_obj.get_nodes())
```

```
: 335829 335811
: 7752240 -1698242034
```

might very well be my bug (in the interface) but wanted to check if you did C++ level tests.

joerowell:
So I didn't do this for the following reasons:

(a) the .at() method provides the checks you'd want, but it throws an
exception: I dont know if we want to enforce the handling that on the
calling code.

(b) the array is quite big, so in the cases we over-read it should return a
0: which seems like the most sensible return value for this.

(c) if we did the checks for this, what would we return? We could return
~uint64_t(0) for invalid, but then we would have to check for that return
type in all places (which would likely screw up the count by quite a bit).


What we could do instead is this: we can use an assert for this (level =<
FPLLL_MAX_ENUM_DIM). That way there won't be a compile time cost if DNDEBUG
is enabled, but for debugging it would flag any potential errors.

Let me know what you think.

joerowell:
So I did run some C++ tests. Its possible I've invoked some compiler
specific behaviour though. I will check it later.

I'm also a bit confused that your count is < 0; the return type is
uint64_t, so this shouldn't be possible.

joerowell:
In any case, I'll add the checks to the PR. We should probably have tests
for this anyway.

malb:
I was casting to a long (now fixed). I think (a) problem is:

```python
In [4]: enum_obj.get_nodes(42)                                                                                                                                                                                                                
Out[4]: 140728307771552

In [5]: enum_obj.get_nodes(41)                                                                                                                                                                                                                
Out[5]: 140728307771536

In [6]: enum_obj.get_nodes(40)                                                                                                                                                                                                                
Out[6]: 0
```
I think `get_nodes(41)` should be zero, right?

joerowell:
> I was casting to a long (now fixed). I think (a) problem is:
> 
> ```python
> In [4]: enum_obj.get_nodes(42)                                                                                                                                                                                                                
> Out[4]: 140728307771552
> 
> In [5]: enum_obj.get_nodes(41)                                                                                                                                                                                                                
> Out[5]: 140728307771536
> 
> In [6]: enum_obj.get_nodes(40)                                                                                                                                                                                                                
> Out[6]: 0
> ```
> 
> I think `get_nodes(41)` should be zero, right?

Yes: this was my fault. It should now be fixed- the ``test_counter.cpp`` file checks for this.

fplll/enum/enumerate.h
joerowell: So I didn't do this for the following reasons:

(a) the .at() method provides the checks you'd want, but it throws an
exception: I dont know if we want to enforce the handling that on the
calling code.

(b) the array is quite big, so in the cases we over-read it should return a
0: which seems like the most sensible return value for this.

(c) if we did the checks for this, what would we return? We could return
~uint64_t(0) for invalid, but then we would have to check for that return
type in all places (which would likely screw up the count by quite a bit).


What we could do instead is this: we can use an assert for this (level =<
FPLLL_MAX_ENUM_DIM). That way there won't be a compile time cost if DNDEBUG
is enabled, but for debugging it would flag any potential errors.

Let me know what you think.

malb: I guess we don't necessarily enforce bounds elsewhere either,e.g. when giving enumeration dimensions, so you're right, we might as well leave it here and push that responsibility on FPyLLL and friends.

malb:
I'm still getting this:

```python
In [1]: import math 
   ...: from fpylll import * 
   ...: FPLLL.set_random_seed(1337) 
   ...: A = LLL.reduction(IntegerMatrix.random(40, "qary", k=20, bits=30)) 
   ...: M = GSO.Mat(A) 
   ...: M.update_gso() 
   ...: pr = Pruning.Pruner(0.99*M.get_r(0,0), 2**20, [M.r()], 0.51) 
   ...: c = pr.optimize_coefficients([1. for _ in range(M.d)]) 
   ...: cost, details = pr.single_enum_cost(c, True) 
   ...: lft = [int(math.ceil(x)) for x in details] 
   ...: enum_obj = Enumeration(M) 
   ...: v = enum_obj.enumerate(0, M.d, 0.99*M.get_r(0,0), 0) 
   ...: rgt = [enum_obj.get_nodes(i) for i in range(M.d)] 
   ...:  
   ...: print(sum(lft), int(math.ceil(cost))) 
   ...: print(sum(rgt), enum_obj.get_nodes(-1)) 
   ...:                                                                                                                                                                                                                                       
335829 335811
7752240 18446744073390048844

```

joerowell:
> I'm still getting this:
> 
> ```python
> In [1]: import math 
>    ...: from fpylll import * 
>    ...: FPLLL.set_random_seed(1337) 
>    ...: A = LLL.reduction(IntegerMatrix.random(40, "qary", k=20, bits=30)) 
>    ...: M = GSO.Mat(A) 
>    ...: M.update_gso() 
>    ...: pr = Pruning.Pruner(0.99*M.get_r(0,0), 2**20, [M.r()], 0.51) 
>    ...: c = pr.optimize_coefficients([1. for _ in range(M.d)]) 
>    ...: cost, details = pr.single_enum_cost(c, True) 
>    ...: lft = [int(math.ceil(x)) for x in details] 
>    ...: enum_obj = Enumeration(M) 
>    ...: v = enum_obj.enumerate(0, M.d, 0.99*M.get_r(0,0), 0) 
>    ...: rgt = [enum_obj.get_nodes(i) for i in range(M.d)] 
>    ...:  
>    ...: print(sum(lft), int(math.ceil(cost))) 
>    ...: print(sum(rgt), enum_obj.get_nodes(-1)) 
>    ...:                                                                                                                                                                                                                                       
> 335829 335811
> 7752240 18446744073390048844
> ```
OK: this means that enumeration is failing (your return value is exactly ~uint64_t(0)). Let me check it's not something on my side (sorry!).

malb:
Sorry, I picked a bad example:

```python
import math
from fpylll import *
d = 44
FPLLL.set_random_seed(1337)
A = LLL.reduction(IntegerMatrix.random(d, "qary", k=d//2, bits=30))
M = GSO.Mat(A)
M.update_gso()
pr = Pruning.Pruner(0.99*M.get_r(0,0), 2**30, [M.r()], 0.51)
c = pr.optimize_coefficients([1. for _ in range(M.d)])
cost, details = pr.single_enum_cost(c, True)
lft = [int(math.ceil(x)) for x in details]
enum_obj = Enumeration(M)
v = enum_obj.enumerate(0, M.d, 0.99*M.get_r(0,0), 0, pruning=c)

rgt = [enum_obj.get_nodes(i) for i in range(M.d)]

print(sum(lft), int(math.ceil(cost)))
print(sum(rgt), enum_obj.get_nodes(-1))
print(enum_obj.get_nodes(d+2))
: 2596847 2596825
: 321613 733988562
: 7187663635689448035


```

malb:
The bug seems fixed now (recompiled and added `fplll_config.h` to FPyLLL explicitly)

malb:
As fas as I am concerned, this can be merged. @cr-marcstevens do you agree?

fplll/enum/enumerate_base.cpp
```diff
@@ -190,7 +190,7 @@ template <bool dualenum, bool findsubsols, bool enable_reset> void EnumerationBa
 
   partdist[k_end] = 0.0;  // needed to make next_pos_up() work properly
 
-  nodes -= k_end - k;
+  nodes[k_end] -= k_end - k;
```
shi-bai, line 193: It seems there could be an overflow? This may affect the following?

      return std::accumulate(nodes.begin(), nodes.end(), 0);

shi-bai:
The function looks useful to me. I have a comment on a potential(?) overflow, with that resolved, the code looks good to me for merging.

fplll/enum/enumerate_base.cpp
joerowell: So I asked Marc about this. This code is apparently to retain compatibility with the loop below it (i.e if we decide to use the iterative enumeration vs the recursive enumeration). Whilst there will be an overflow, the recursive dispatch accounts for this, and so it will still return the correct count.

shi-bai: I am puzzled. Just tried the test_counter.cpp. I added a line
    cout << "Level " << i << " counts " << a[i] << endl;
right after Line 28. 

For the first (10 + 1) levels, it prints:

Level 0 counts 1
Level 1 counts 1
Level 2 counts 1
Level 3 counts 1
Level 4 counts 1
Level 5 counts 1
Level 6 counts 4
Level 7 counts 4
Level 8 counts 3
Level 9 counts 2
Level 10 counts 18446744073709551606

Do you mean the Level 9 #nodes should be 12 instead of 2?

cr-marcstevens:
The correct behaviour for this correction would be to subtract 1 from each level so corrected, instead of correcting N levels by subtracting N nodes at the toplevel.
However, maybe this is just a troublesome correction (and might confuse future fplll developers) and we could try to just go without.

fplll/enum/enumerate_base.cpp
shi-bai: Marc & Joe, the maximum mis-counting is by a value of k? If so, this is minor in practice for larger k.

joerowell: Whilst that's true, I think we should make sure that all of the enumerators return the same number of nodes. Otherwise, we could end up with a scenario in future where something needs to be debugged, and if the number of nodes is significantly different then that may cause some headaches. Of course for large k this 'significantly different' will be small compared to the overall count, but I'm still concerned about it in practice for small values of k.

joerowell:
> The correct behaviour for this correction would be to subtract 1 from each level so corrected, instead of correcting N levels by subtracting N nodes at the toplevel.

I'm not sure I agree with this. If we subtracted 1 from each level that's been visited, then presumably for small enumerations (i.e say over dimension 10) we'd end up with a 0 count at the bottom-most node, and so the count would imply that we didn't find any solutions. Or am I misunderstanding?

cr-marcstevens:
That's not what I meant.
Instead of starting at the input level k, the code changes this to kend-1 so recursive enumeration might be called.
This adds additional nodes to the initial descend of the enumeration: 1 additional node at each level k+1, ..., kend-1.
But currently these additional nodes are all corrected at level kend-1, instead of their own level.

joerowell:
Oh that makes more sense; sorry for the misunderstanding.

shi-bai:
Thanks for the clarification, Marc and Joe. The current PR fa9ba6b looks good to me. Joe, please merge to main.

cr-marcstevens:
+1

joerowell:
> Thanks for the clarification, Marc and Joe. The current PR [fa9ba6b](https://github.com/fplll/fplll/commit/fa9ba6b9d0beb7544200f9743d4945639ed1c95e) looks good to me. Joe, please merge to main.

I don't have write access, is this something you can do for me?

Thanks all for the helpful comments and feedback!

PR #447, opened 2020-10-17
Enumeration Algorithm on GPU using CUDA

FeanorTheElf:
Pull Request to integrate an existing implementation of the enumeration algorithm for GPUs into fplll. The algorithm is written using the CUDA technology by Nvidia. This should lead to faster enumeration when corresponding GPUs are present, as the enumeration is parallelizable well.

Currently, this is work in progress, as the existing implementation requires heavy refactoring. At the moment, the code should be compile-and executable on its own, but is not integrated into fplll and there are also some fundamental parts missing.


cr-marcstevens:
Hi FeanorTheElf, I haven't had a deep look yet, but a quick question:
Does your library use the external-enumeration-library API (https://github.com/fplll/fplll/blob/master/fplll/enum/enumerate_ext.h)?
Or if not, is there another connection so that fplll enumeration may call your CUDA code?

Thanks!
Marc

FeanorTheElf:
I have not built such a connection yet. The problem is, that the external enumeration API requires passing callbacks (as std::functions), which will not be callable from the GPU, as code for the GPU is compiled in a completely different way. I am unsure which is the best way to work around this, but probably we will have to get rid of the callbacks or compile them also for the GPU.

FeanorTheElf:
The problem seems to be quite hard to solve, as the `process_(sub)solution()` callback has to be called often during the enumeration, which would normally require it to be executable on the gpu (not just in the external enumeration API, as otherwise, one must call `process_sol()` on an Evaluator object). Being callable with different evaluators therefore requires any kind of function pointer idea. However, this does not work because of a fundamental caveat in cuda (https://forums.developer.nvidia.com/t/consistency-of-functions-pointer/29325).

I only see the two following alternatives:
- A queue passing solutions to the cpu and results back to the device, using system atomics
- Determine evaluators statically via templates
The first solution is likely to introduce a lot of overhead and is hard to implement (and requires system atomics, which are not available on all devices). Therefore, I have now implemented the second solution, with the caveat that the cuda enumeration only supports a fixed set of possible evaluators. If someone has a better idea, please tell me.

FeanorTheElf:
This pull request is obsolete, as its functionality will be completely provided by https://github.com/fplll/fplll/pull/456.

cr-marcstevens:
Hi Simon,

The CUDA code has largely been removed into a separate repo that can be used together with fplll.
Indeed better to remove this pull request and to:
- create an issue with tag enhancement and link to your new CUDA enum library that works with fplll.
- to create a new pull request updating fplll's README with link to your library and instructions how to use it.

PR #450, opened 2020-11-17
Added files implementing cvpp functionality within the sieving framew…

ManolisDoulgerakis:
This pull request includes an implementation of the CVPP functionality. It is a slightly updated version of the implementation used for the CVPP part of the experiments included in the paper: 
Sieve, Enumerate, Slice, and Lift: Hybrid Lattice Algorithms for SVP via CVPP. 

The main part of the new code is included in the file sieve_cvpp.cpp. It implements the iterative slicer and the randomised slicer. The implementation is constructed in the rational that the user gives at least two files as input, one which includes a basis of the lattice and one which includes the target vectors. The list of vectors associated with the preprocessing can either be given as input from a file or can be constructed by the already implemented sieving algorithms. In case the later option is chosen the user can select if he wants the preprocessed list to be stored in a file for later use. Finally the output of the randomised slicer is stored in a file whose name is constructed based on the name of the file containing the target vectors. For example if the target vectors were stored in a file targets.txt the output of the algorithm will be stored in targets-cvpp_solutions.txt

An example call to solve a CVPP would be the following:
latsieve -f basis.txt -L preprocessed_list.txt -C targets1000.txt -v

As the implementation was made within the GaussSieve class the sieve_main.cpp file was also modified in order for the main call to recognise when it is asked to solve SVP or CVPP.

### DIFF
--- a/fplll/sieve/sieve_cvpp.cpp
+++ b/fplll/sieve/sieve_cvpp.cpp
@@ -1,0 +1,316 @@
+#include "sieve_gauss.h"
+
+/**
+ * sets the filename for the output
+ */
+string solutions_filename(const char *targets_file_name)
+{
+  string file_name = string(targets_file_name);
+  string old_name;
+  old_name = file_name.substr(0, file_name.find_last_of('.'));
+  return old_name + string("-cvpp_solutions.txt");
+}
+
+/**
+ * imports target vectors from file
+ * Note: the targets are imported without their norm included.
+ */
+template <class ZT> list<ListPoint<ZT> *> target_import(const char *targets_file_name, const int nc)
+{
+  list<ListPoint<ZT> *> target_list;
+  NumVect<Z_NR<ZT>> t_vec(nc);
+  ListPoint<ZT> *p;
+
+  ifstream input_file(targets_file_name);
+  if (!(input_file.is_open()))
+  {
+    cout << "# [ERROR] The input file " << targets_file_name;
+    cout << " was not opened properly. The target(s) were not imported." << endl;
+  }
+  else
+  {
+    p = new_listpoint<ZT>(nc);
+    while (input_file >> t_vec)
+    {
+      p = num_vec_to_list_point(t_vec, nc);
+      if (p->norm > 0)
+        target_list.push_back(p);
+    }
+    input_file.close();
+  }
+  return target_list;
+}
+
+/**
+ * imports List from a file
+ * Note: the list of vectors is imported without their norms
+ */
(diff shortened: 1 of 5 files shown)


malb:
Proposal, we do the removal of the gauss sieve first and then merge this PR into the gauss-sieve branch

ManolisDoulgerakis:
If that helps in the process of merging then it is no problem from my side.

malb:
Can you send a PR against this branch instead? https://github.com/fplll/fplll/tree/gauss-sieve

ManolisDoulgerakis:
OK. Is it possible to do it through this PR or should I close this one and open a new one?

ManolisDoulgerakis:
I tried to modify this PR. Did it work?

joerowell:
Hi Manos,

I'm not requested to review this, but I have a few comments. None of these comments are on correctness (I haven't look that deeply), but I noticed a couple of things. 

- Is there a reason you're using ```std::list``` over ```std::vector```?  To my understanding, there's a significant performance degradation for using ```std::list``` over ```std::vector``` because the elements are not contiguous in memory and so caching doesn't really have a chance to shine.

- I've noticed that you've got some calls to ```new``` and ```delete``` directly. Is there a reason for this? It's typically common in 'Modern C++' to minimise the usage of ```new``` and ```delete``` when possible, but to instead take advantage of features like rvalue semantics and similar to achieve the right meaning. The reason for this is then the destructors of each object will automatically get called when the object falls out of scope - in other words, the freeing is deterministic and you don't need to worry about it. You could also subsume everything you have in the ```new_listpoint``` function into the constructor for ```ListPoint``` and not lose any functionality.

Happy to discuss any of this further if that helps :)

ManolisDoulgerakis:
Hi Joe,

Thanks for your comments! 
Well, more or less my answer to both of your observations is the same. 
I followed the style in the initial implementation of the sieve. In order to handle
the list of sieving it is used `std::list` so I thought I should follow the same style 
used in the implementation I was trying to "extend". 

The same holds about the functions `new_listpoint()` and `del_listpoint()` if you check 
for example the file sieve_gauss_4sieve.cpp calls to these two functions are all over the place.

At least from my side, I would wish my code to be used mainly as an educational tool, so I am 
not aiming so much on performance here. So, somebody who wants to start studying about sieving
and CVPP can have a nice-and-clear hands-on experience. The fact that this code is not anymore in 
the master branch I think that supports this point of view.

What do you think?

P.S. I am sorry for the late reply.

Issue #244, opened 2017-04-26
usability of `infinite loop in babai`


malb:
The error message `infinite loop in babai` causes confusion amongst users, cf. https://github.com/fplll/fplll/issues/170#issuecomment-297404695 I propose:

1. We create an FAQ where we explain what that error message means and propose to increase precision (a second item could then explain how to do that depending on situation)

2. Add a link to the FAQ entry to the error message.

joerowell:
I think we can symbolically close this issue (since #453 )

PR #453, opened 2020-11-19
fplll faq page

joerowell:
We've all had the "infinite loop in Babai" error code. But what does it mean?

## This PR:
This PR attempts to make this a bit nicer. In particular, we now have a wiki page (https://github.com/fplll/fplll/wiki/fplll-errors-FAQ) that is now printed to the user whenever an error message is printed. This is meant to be a solution to #244 .  

## What does it need?
Please check the error codes/recommended solutions on the wiki page. I've only briefly encountered some of these errors before, and so it isn't likely to be 100% accurate.

### DIFF
--- a/fplll/defs.h
+++ b/fplll/defs.h
@@ -168,19 +168,20 @@ enum RedStatus
   RED_STATUS_MAX        = 13
 };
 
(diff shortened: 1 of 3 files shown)


malb:
This looks good to me.

PR #455, opened 2020-11-19
Installation from package instructions (fixes #190) 

### LINKED ISSUE #190: README: differentiate building from source and using packaging tools
For information, fplll is now available on macOS via [homebrew](http://brew.sh) (third-party packaging tool) 

`brew install homebrew/science/fplll`

(see [fplll.rb on homebrew-science](https://github.com/Homebrew/homebrew-science/blob/master/fplll.rb))

I think the README.md file could be slightly reorganized. Indeed, the dependencies are only when building from GitHub (one does not need autotools when building from the release and when using fplll). It could be nice to have the dependencies for using the libraries and ways to install it from 3rd party tools (TODO: make ubuntu / debian / archlinux packages), and an explanation on how to build it from source with autotools as dependency.

joerowell:
We can install fplll in multiple ways. Shouldn't the readme talk about this?

## This PR:
This PR attempts to close a long-standing issue (#190) that pertains to describing some different ways to install fplll from pre-built packages. In particular, this PR describes how to install fplll from packages on the following platforms:

- Ubuntu and Debian
- MacOS
- Conda
- Docker.

It also makes some small wording changes to the WSL section (just small tweaks to my own wording from fplll days 5). This should make the readme clearer.

## What needs to be done?
I've checked all of the instructions but the Docker ones (this simply refers the user to the repository with the Docker images); these all work . 

It's possible though that I've screwed something up, and I'd appreciate it if someone could check at least some subset of these instructions!

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ fplll is available as a pre-built package for a variety of operating systems; th
 
 Below, we give some instructions on how to install these packaged variants of fplll.
 
-Note that these packages will be up-to-date for the most recent version of fplll. However, if you want a feature that has recently been added to master (that is not yet in a release) then it is necessary to build from source (See the "Install from Source" section).
+Note that these packages will be up-to-date for the most recent version of fplll. However, if you want a feature that has recently been added to master (that is not yet in a release) then it is necessary to build from source. If this is the case, please see the [Installation from Source](#installation-from-source) section.
 
 ### Ubuntu and Debian ###
 fplll can be installed directly via Aptitude or Synaptic. Both of these package managers package fplll in the package `fplll-tools`. Therefore, to install this package using Aptitude, run the following command
--- a/README.md
+++ b/README.md
@@ -325,8 +325,9 @@ The options are:
 See [API documentation](https://fplll.github.io/fplll/annotated.html) and [tests](https://github.com/fplll/fplll/tree/master/tests) as a source of examples.
 
 ## Multicore support ##
+This library does not currently use multiple cores and running multiple threads working on the same object `IntegerMatrix`, `LLLReduction`, `MatGSO` etc. is not supported.  Running multiple threads working on *different* objects, however, is supported. That is, there are no global variables and it is safe to e.g. reduce several lattices in parallel in the same process.
 
-This library does not currently use multiple cores and running multiple threads working on the same object `IntegerMatrix`, `LLLReduction`, `MatGSO` etc. is not supported. Running multiple threads working on *different* objects, however, is supported. That is, there are no global variables and it is safe to e.g. reduce several lattices in parallel in the same process.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,27 +57,65 @@ fplll is distributed under the [GNU Lesser General Public License](COPYING) (eit
...
 # Compilation #
 
-## Dependencies ##
+## Installation from packages ##
+
+fplll is available as a pre-built package for a variety of operating systems; these pre-built packages typically include all mandatory dependencies, and so these packages can be used to start running fplll quickly.
+
+Below, we give some instructions on how to install these packaged variants of fplll.
+
+Note that these packages are likely to be rebuilt every time a new tag is cut; as a result, if you want fully up-to-date code, then it is necessary to build from source (See the "Install from Source" section).
```
malb, line 66: I don't follow the logic here.

README.md
```diff
@@ -52,27 +57,65 @@ fplll is distributed under the [GNU Lesser General Public License](COPYING) (eit
...
+### MacOS ###
+
+MacOS has a package for fplll inside HomeBrew. Assuming that you have HomeBrew installed, you may install fplll using the following command
+
+```
+brew install fplll
+```
+
+### Docker and AWS ###
+We now have Docker/AWS images for fplll too. They aren't on this repository, though; you can find them [here](https://github.com/malb/fplll-virtual-machines)
```
malb, line 94: I'm in the process of changing that up better to link to https://hub.docker.com/u/fplll

README.md
```diff
@@ -52,27 +57,65 @@ fplll is distributed under the [GNU Lesser General Public License](COPYING) (eit
...
+
+#### Required ####
 
 - GNU MP 4.2.0 or higher [http://gmplib.org/](http://gmplib.org/) or MPIR 1.0.0 or higher [http://mpir.org](http://mpir.org)
 - MPFR 2.3.0 or higher, COMPLETE INSTALLATION [http://www.mpfr.org/](http://www.mpfr.org/)
 - autotools 2.61 or higher
 - g++ 4.9.3 or higher
 
-### Optional ###
+#### Optional ####
```
malb, line 109: I see that the below isn't actually new, but the Windows 10 instructions are confusing. The user does need those pre-requirements.

malb (commented):
Nice, only have some minor comments

README.md
joerowell: I think I've fixed this.

README.md
joerowell: Fixed (I think)

README.md
```diff
@@ -282,8 +325,9 @@ The options are:
...
 
-This library does not currently use multiple cores and running multiple threads working on the same object `IntegerMatrix`, `LLLReduction`, `MatGSO` etc. is not supported. Running multiple threads working on *different* objects, however, is supported. That is, there are no global variables and it is safe to e.g. reduce several lattices in parallel in the same process.
+As an exception to the above, fplll has an implementation of parallel lattice point enumeration. This implementation is currently used as the default enumeration strategy in fplll. At present fplll does not contain strategies for multi-core pruned enumeration, and so speedups for pruned enumeration may be sub-linear (see [this](https://martinralbrecht.wordpress.com/2020/10/26/multicore-bkz-in-fplll/) for more information). On the other hand, unpruned enumeration appears to scale linearly.
```
malb, line 330: We actually set single thread as the default.

PR #456, opened 2020-11-20
Set a default external enumeration library

cr-marcstevens:
This PR makes it possible to directly integrate any external enumeration into fplll and be used by default by fplll.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -188,7 +188,7 @@ AM_CONDITIONAL(FPLLL_PARALLEL_ENUM, test "x$max_parallel_enum_dim" != "x0")
 # Use external enumeration
 AC_ARG_WITH(extenum-dir,
   AS_HELP_STRING([--with-extenum-dir@<:@=DIR@:>@], [Specify external enumeration library install directory]),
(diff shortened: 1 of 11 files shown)


malb:
Looks good to me, so once the checkstyle issue is fixed, happy to merge.

cr-marcstevens:
Thanks, but indeed all checks failed, so have to fix style and maybe other things.

cr-marcstevens:
BTW this pull request moves the extenum API from the fplll namespace to the main namespace `::`.
This is because the header file doesn't use any other fplll header files, so that it can be copied to extenum libraries without dependencies.

What are your thoughts on this:
1. Should this be moved back to fplll? 
1a. If yes, do we use macros then or just hardcoded `fplll` as namespace?

malb:
TBH, I don't have a strong opinion either way. What's the risk of name collision in some bigger project, i.e. should it be in some namespace?

cr-marcstevens:
@malb do you want to push the green button?

malb:
With pleasure.

PR #458, opened 2020-12-02
fix enum ext callback issue

### LINKED ISSUE #449: External Enumeration `callback_set_config()` loop index bug
The function `callback_set_config()` in the class `ExternalEnumeration` (that is also passed to external enumerators as callback) throws an assertion error (in Debug builds) when called with `mutranspose == false`.

The reason is the following code:
```

  // Now we copy the mu values from the gso matrix.
  if (mutranspose)
  {

    size_t offs = 0;
    for (int i = 0; i < _d; ++i, offs += mudim)
    {
      for (int j = i + 1; j < _d; ++j)
      {
        _gso.get_mu(fmu, j + _first, i + _first);
        /* mu[i][j]= */
        mu[offs + j] = fmu.get_d();
      }
    }
  }
  else
  {
    size_t offs = 0;
    for (int j = 0; j < _d; ++j, offs += mudim)
    {
      for (int i = 0; i < _d; ++i)
      {
        _gso.get_mu(fmu, j + _first, i + _first);
        /* mu[j][i] = */
        mu[offs + i] = fmu.get_d();
      }
    }
  }
```
In the case `mutranspose == true`, it correctly calls `_gso.get_mu(i, j)` only for i > j, but in the other case, this condition is not satisfied, triggering the assertion (note that i and j are swapped here)
```
MatGSOInterface<ZT, FT>::get_mu(FT &f, int i, int j)
{
  FPLLL_DEBUG_CHECK(i >= 0 && i < n_known_rows && j >= 0 && j < i && j < gso_valid_cols[i] && !in_row_op_range(i));
  ...
```
in `MatGSOInterface<ZT, FT>::get_mu()`. Fixing this should be pretty easy, and I would be willing to do so in my pull request https://github.com/fplll/fplll/pull/447 (where I also encountered the problem). However, this would of course tie the fix to merging the Cuda Enumeration.

cr-marcstevens:
Fixes #449

### DIFF
--- a/fplll/enum/enumerate_ext.cpp
+++ b/fplll/enum/enumerate_ext.cpp
@@ -113,7 +113,7 @@ void ExternalEnumeration<ZT, FT>::callback_set_config(enumf *mu, size_t mudim, b
     {
       for (int j = i + 1; j < _d; ++j)
       {
-        _gso.get_mu(fmu, j + _first, i + _first);
+        _gso.get_mu(fmu, j + _first, i + _first);  // j > i
         /* mu[i][j]= */
         mu[offs + j] = fmu.get_d();
       }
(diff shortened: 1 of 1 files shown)


cr-marcstevens:
@malb ready to merge.

Issue #464, opened 2021-05-21
`shortest_vector` does not return shortest vector


malb:
Consider this experiment:

```python
from fpylll import *
from math import sqrt

def norm(v):
    n = 0
    for v_ in v:
        n += v_**2
    return sqrt(n)

FPLLL.set_random_seed(0)
A = LLL.reduction(IntegerMatrix.random(180, "qary", k=90, q=65537))
print(norm(A[0]))
print(norm(SVP.shortest_vector(A[:41], method="proved", preprocess=2)))
print(norm(SVP.shortest_vector(A[:42], method="proved", preprocess=2)))
print(norm(SVP.shortest_vector(A[:43], method="proved", preprocess=2)))
```

We would expect the vector found for A[:42] not to be longer than the one found for A[:41] but we get:

```
12633.474066938199 # LLL
11629.795441021308 # A[:41]
11678.886762016318 # A[:42]
10459.696936336157 # A[:43]
```

malb:
C++version

```c++
#include <fplll/fplll.h>
#include <math.h>

using namespace fplll;

void submatrix(ZZ_mat<mpz_t> &T, ZZ_mat<mpz_t> &F, size_t a, size_t b)
{
  T.resize(b - a, F.get_cols());
  for (size_t i = 0; i < (b - a); i++)
  {
    for (size_t j = 0; j < F.get_cols(); j++)
      T[i][j] = F[a + i][j];
  }
}

double norm(const vector<Z_NR<mpz_t>> &v)
{
  Z_NR<mpz_t> norm, tmp;
  for (size_t i = 0; i < v.size(); i++)
  {
    tmp.mul(v[i], v[i]);
    norm.add(norm, tmp);
  }
  return sqrt(norm.get_d());
}

int main(int argc, char *argv[])
{
  ZZ_mat<mpz_t> A(180, 180);
  A.gen_qary_withq(90, 65537);

  lll_reduction(A);
  ZZ_mat<mpz_t> B, C, D;
  vector<Z_NR<mpz_t>> b_sol_coord, c_sol_coord, d_sol_coord;
  vector<Z_NR<mpz_t>> b_solution, c_solution, d_solution;
  submatrix(B, A, 0, 41);
  submatrix(C, A, 0, 42);
  submatrix(D, A, 0, 43);

  shortest_vector(B, b_sol_coord);
  shortest_vector(C, c_sol_coord);
  shortest_vector(D, d_sol_coord);

[...]

PR #466, opened 2021-05-22
Exacteval fix

cr-marcstevens:
Should fix issue #464

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,4 @@ doc/html/
 /tests/test_enum
 /tests/test_bkz_gram
 /tests/test_svp_gram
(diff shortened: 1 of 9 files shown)


cr-marcstevens:
Note that test_svp doesn't actually check anything with u.

malb:
> Note that test_svp doesn't actually check anything with u.

Yeah, but it tests (a) that you can do this multiplication (which incidentally failed) and documents that this is how you get the solution wrt to the original basis. Adding regression tests now.

malb:
If you're happy with my additions and tests pass, I'd say this is ready to be merged. Thanks Marc!

Issue #467, opened 2021-06-14
CVP Enumeration seems bugged


lducas:
I'm playing with the Leech lattice, and I'm noting incorrect results:
- the Leech lattice has exactly 196560 shortest vectors, of square norm 32 (for this particular scaling)

I set the enumeration radius at 32.5 to handle small errors.

If I Enumerate without target, I get half of that (expected behavior as SVP enumeration only search in half the space).
If I enumerate with target=0, I still only get half of that, which is not the behaviour I'd expect. 
If I enumerate with something very close to 0, I get a very strange number: 144856. More than half, but still not whole of it :/

(sorry, my test case is using fpylll)

``` python
from fpylll import IntegerMatrix, Enumeration, LLL, GSO

LeechBasis = [[8, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[4, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[4, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[4, 0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[4, 0, 0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[4, 0, 0, 0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[4, 0, 0, 0, 0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[2, 2, 2, 2, 2, 2, 2, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[4, 0, 0, 0, 0, 0, 0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
[...]

PR #468, opened 2021-06-14
bugfix CVP enumeration

lducas:
This is a partial fix for issue #467 . It only fixes it for enumerate_base, that is when we use flags `--disable-recursive-enum --with-max-parallel-enum-dim=0`


cr-marcstevens:
Note that there are 3 enumeration implementations in FPLLL:

1. loop: a generic implementation that always works.
2. recursive: a special compiled version for each enumeration dimension.
3. extenum/enumlib: external enumeration API in combination with a default parallel enumeration library. This only supports SVP

The implementation selection goes as follows: 

1. extenum if: enumlib enabled & it is SVP (not CVP) & it is not subtree enum & the enumeration call does not get rejected by the library (e.g. dimension is not compiled)
2. recursive if: enabled and compiled for that dimension
3. loop: always works

Hence to properly test enumeration implementations, our CI tests should cover versions:
- recursive and extenum enabled
- recursive enabled, extenum disabled
- recursive & extenum disabled

cr-marcstevens:
The external enumeration library can be disabled at runtime, which could be covered by tests (but likely isn't now).
However recursive enumeration can only be disabled at compile time.
Enabling recursive & extenum can also be done via configure:

1. `./configure --enable-recursive-enum --max-parallel-enum-dim=80`
2. `./configure --enable-recursive-enum --max-parallel-enum-dim=0`
3. `./configure --disable-recursive-enum --max-parallel-enum-dim=0`

malb:
I think adding the CI plumbing shouldn't be hard and I can take care of that. Proposal: you add a test to `test_svp` and `test_cvp`. I then make sure we cover 

1.  `./configure --enable-recursive-enum --max-parallel-enum-dim=80`
2.  `./configure --enable-recursive-enum --max-parallel-enum-dim=0`
3.  `./configure --disable-recursive-enum --max-parallel-enum-dim=0`

I'd just copy-n-paste `tests.yml`.

lducas:
Ok, test done. (Though apparently my check-style doesn't match the CI check-style). You can do your magic @malb  !

lducas:
Thanks Marc ! I'd never have figured out the dist-check...

malb:
All done on my end. It's still marked as "draft", is it ready to merge?

lducas:
I'm happy with it as well. @cr-marcstevens ?

PR #472, opened 2021-07-15
Make some methods in MatGSO const

joerowell:
# What's the problem?

At the moment I can't write this code:

``` 
template<typename Z, typename F>
void do_something(const MatGSO<Z_NR<Z>, FP_NR<F>>& gso) {
       const auto r = gso.get_r_matrix();
}
```
Because ```get_r_matrix``` isn't a const function, even though it returns a const reference:

```
  const Matrix<FT> &get_r_matrix() { return r; }
```

This PR fixes this by denoting this method (and I believe all others) in ```gso_interface.h``` that return const references as ```const```. Thus, the above method now reads as:

```
  const Matrix<FT> &get_r_matrix() const { return r; }
``` 

Which enables the code snippet to compile.

### DIFF
--- a/fplll/gso_interface.h
+++ b/fplll/gso_interface.h
@@ -202,21 +202,21 @@ public:
    * (lower triangular matrix)
    * mu(i, j) = r(i, j) / ||b*_j||^2.
    */
-  const Matrix<FT> &get_mu_matrix() { return mu; }
+  const Matrix<FT> &get_mu_matrix() const { return mu; }
 
   /**
    * Returns the r matrix
    * Coefficients of the Gram Schmidt Orthogonalization
    * (lower triangular matrix)
    */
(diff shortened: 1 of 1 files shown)


joerowell:
There's probably other functions in fplll that do this. If it's considered important I'm happy to fix those too, but it might take a little longer.

malb:
Sorry, I'm daft: why is the above code snippet bad?

joerowell:
> Sorry, I'm daft: why is the above code snippet bad?

Do you mean why is it bad that I can't write that code snippet?

malb:
> > Sorry, I'm daft: why is the above code snippet bad?
> 
> Do you mean why is it bad that I can't write that code snippet?

"Can't"!! I'm an idiot. Ignore.

PR #473, opened 2021-07-15
Fix #470 (document the role of u in LLL_reduction)

### LINKED ISSUE #470: document role of `U` in `lll_reduction`
That is, we overwrite it with the transformation matrix.

joerowell:
This PR is an attempt to fix #470.

The simplest way (I think) to do this is to add some documentation to the class. I believe what I've written is correct, but another pair of eyes would be appreciated.

### DIFF
--- a/fplll/wrapper.h
+++ b/fplll/wrapper.h
@@ -32,11 +32,17 @@ FPLLL_BEGIN_NAMESPACE
  * heuristic and fast variants of both LLL and HLLL.
  *
  * User note: all of the parameters passed into this class are non-const references. Thus, this
- * class may overwrite these variables. In particular, the methods in this class typically
+ * class may overwrite these variables. 
+ *
+ * In particular, the methods in this class typically
  * over-write the parameter ``u`` with the transformation matrix applied to ``b`` to produce LLL(B).
  * In other words, let C = LLL(B). Then we can write C = U B for some unimodular transformation
  * matrix U. For this reason, ``u`` should either be an empty matrix or the identity matrix when it
- * is passed in to this function, so that no information is lost.
+ * is passed in to this function, so that no information is lost. 
+ *
+ * Similarly, the parameter ``u_inv`` will contain the inverse matrix ``u ^ -1``. 
+ * This allows you to recover the original basis ``b`` by multipyling ``C`` by ``u_inv``.
+ * This and other such behaviour on these parameters is described in some detail in gso_interface.h. 
  */
 class Wrapper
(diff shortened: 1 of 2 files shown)


fplll/wrapper.h
```diff
@@ -25,10 +25,23 @@ FPLLL_BEGIN_NAMESPACE
...
+ * over-write the parameter ``u`` with the transformation matrix applied to ``b`` to produce LLL(B).
+ * In other words, let C = LLL(B). Then we can write C = U^T B for some unimodular transformation
+ * matrix U. For this reason, ``u`` should either be an empty matrix or the identity matrix when it
+ * is passed in to this function, so that no information is lost.
+ */
 class Wrapper
 {
 public:
-  /* u must be either empty or the identity matrix */
+  /* u must be either empty of the identity matrix */
```
malb, line 44: typo

fplll/wrapper.h
```diff
@@ -25,10 +25,23 @@ FPLLL_BEGIN_NAMESPACE
 /* The matrix b must not be modified before calling lll().
    lll() must be called only once. */
 
+/**
+ * @brief Wrapper. This class provides an externally callable API for LLL reducing some basis ``b``.
+ * This class forcibly instantiates some template declarations (see FPLLL_DECLARE_LLL(T) for more
+ * information at the bottom of this class) and provides an interface for calling provable,
+ * heuristic and fast variants of both LLL and HLLL.
+ *
+ * User note: all of the parameters passed into this class are non-const references. Thus, this
```
malb, line 34: It's not the transpose:
```
sage: from fpylll import *                                                                                                                                                                                                                    
sage: A = IntegerMatrix.random(10, "qary", k=5, q=127)                                                                                                                                                                                        
sage: U = IntegerMatrix(10, 10)                                                                                                                                                                                                               
sage: B = LLL.reduction(copy(A), U=U)                                                                                                                                                                                                         
sage: U.to_matrix(matrix(ZZ, 10, 10)) * A.to_matrix(matrix(ZZ, 10, 10)) == B.to_matrix(matrix(ZZ, 10, 10))                                                                                                                                    
True
```

We don't have matrix-matrix products in FPyLLL so I'm using Sage for that.

joerowell: I wrote ``U^T`` because we use row-representation for both ``A`` and ``U``, no? Otherwise we'd be multiplying a column by a row when we want to multiply row by row. (It's very possible I'm mis-understanding something here!)

Issue #477, opened 2021-09-17
Class name collision with fplll.


lingerois:
I have defined my own Matrix class. However, if I add
```
#include <fplll.h>
```
into my work, it has name collision with fplll.

I think this might caused by there's code like
```
using namespace fplll;
```
In the fplll library itself.

I hope you could help to figure out the right way to use fplll library in my C++ work.

cr-marcstevens:
Could you point to that code `using namespace fplll` ?

BTW there is also a workaround for ambiguations by specifying the namespace where needed:
E.g. `my_namespace::Matrix` and `fplll::Matrix` instead of `Matrix`

lingerois:
> Could you point to that code `using namespace fplll` ?
> 
> BTW there is also a workaround for ambiguations by specifying the namespace where needed:
> E.g. `my_namespace::Matrix` and `fplll::Matrix` instead of `Matrix`

Thanks. After precompile, I found code in the output file shown as following:
```
# 34 "/usr/local/include/fplll/fplll.h" 2 3
# 2 "/usr/local/include/fplll.h" 2 3
using namespace fplll;
```

Howevery, I know fplll.h doesn't contain such line of code. So this is what confused me. It might also be worthy to mention that I am using the release 5.4.1 version of fplll.

lingerois:
The precompile translate
```C++
#include <fplll.h>
```
to the attached file
[out.txt](https://github.com/fplll/fplll/files/7185597/out.txt)
you can read out
```C++
using namespace fplll;
```
from line `85525`.

joerowell:
I think this comes from this line:
https://github.com/fplll/fplll/blob/37bcab0caf510dd7ad49d8a3b277366d7b8b763a/fplll/Makefile.am#L127

I'm at a loss as for why this happens though. Maybe @malb knows?

malb:
This seems bad and we should remove it, right?

joerowell:
This should now be fixed. @lingerois could you please clone and install the most recent version of the repo and see if that helps? We'll probably have to release a new version to fix this for ```apt``` packages and similar, but that's above my pay-grade.

lingerois:
Thanks for you effort! @joerowell It seems nice and well now :)

PR #478, opened 2021-09-17
Remove 'using namespace fplll' (fix #447)

joerowell:
This PR just removes the "using namespace fplll" line from  ```fplll/Makefile.am ``` due to the ticket #477.

Compulation, installation and make check all work locally. I also tested this by pre-processing the following test file:

```
#include<iostream>
#include<fplll.h>

int main() {
    ZZ_mat<mpz_t> mat;
    std::cin >> mat;
    std::cout << mat ;
   return 0;
}

```
Running ```gcc -E``` on this gives the following (explicitly shortened):

```
} 
# 34 "/usr/local/include/fplll/fplll.h" 2 3
# 1 "/usr/local/include/fplll.h" 2 3
# 3 "test.cpp" 2
```

In other words, the ```using namespace fplll``` is no longer present.

### DIFF
--- a/fplll/Makefile.am
+++ b/fplll/Makefile.am
@@ -124,7 +124,6 @@ libfpllld_la_LDFLAGS=$(libfplll_la_LDFLAGS)
 
 install-data-hook:
 	echo "#include <fplll/fplll.h>" > "$(builddir)/fplll.h.root"
-	echo "using namespace fplll;" >> "$(builddir)/fplll.h.root"
 	$(INSTALL) -m 644 "$(builddir)/fplll.h.root" "${DESTDIR}$(includedir)/fplll.h"
 	rm -f "$(builddir)/fplll.h.root"
 


joerowell:
I kept the two separate fplll files: this PR _just_ removes the ``` using namespace fplll;```. 
I don't think it's objectively bad to create this sort of wrapper: it makes it easier to include the library elsewhere.

malb:
LGTM. Thanks! Can I check, compiling your test programme would now fail but previously would not, right?

joerowell:
> LGTM. Thanks! Can I check, compiling your test programme would now fail but previously would not, right?

Yes, exactly. With this PR, you'll get the following error:

```
test.cpp:6:5: error: ‘ZZ_mat’ was not declared in this scope; did you mean ‘fplll::ZZ_mat’?
    6 |     ZZ_mat<mpz_t> mat;
      |     ^~~~~~
      |     fplll::ZZ_mat

```

Which is what we need to fix #477.

PR #480, opened 2021-09-23
Fix for #479 - replace static mpfr with temporaries

### LINKED ISSUE #479: potential issue in multi-thread mode
I have an issue when I use lll_reduction within the bacsel tool (https://gitlab.inria.fr/zimmerma/bacsel) in multi-thread mode. The issue disappears with only one thread, or when I use LM_PROVED instead of LM_WRAPPER, which makes me think it is related to fplll. To reproduce, use revision b7f6b0d of bacsel:
```
make GMP=... MPFR=... DEFSAL="-DACOS -DBASIS=2 -DAUTOMATIC"
./bacsel -rnd_mode all -prec 128 -n 53 -nn 53 -m 43 -t 19.6 -t0 4503599627370496 -t1 9007199254740992 -d 2 -alpha 2 -v -e_in -9 -nthreads 64
...
Dimensions of the lattices: 6x9
../../src/get_ld.c:333: MPFR assertion failed: (ret >= 0.5 && ret < 1.0) || (ret <= -0.5 && ret > -1.0)
```
gdb says:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff6ba5537 in __GI_abort () at abort.c:79
#2  0x00007ffff6bfe768 in __libc_message (action=action@entry=do_abort,
    fmt=fmt@entry=0x7ffff6d0ce2d "%s\n") at ../sysdeps/posix/libc_fatal.c:155
#3  0x00007ffff6c05a5a in malloc_printerr (
    str=str@entry=0x7ffff6d0f1c8 "double free or corruption (fasttop)")
    at malloc.c:5347
#4  0x00007ffff6c06d55 in _int_free (av=0x7ffec4000020, p=0x7ffec4001cc0,
    have_lock=0) at malloc.c:4266
#5  0x00007ffff714e299 in mpfr_free_func ()
   from /lib/x86_64-linux-gnu/libmpfr.so.6
#6  0x00007ffff7145f0d in mpfr_clear () from /lib/x86_64-linux-gnu/libmpfr.so.6
#7  0x00007ffff77face0 in int fplll::Wrapper::call_lll<__mpz_struct [1], double>(fplll::ZZ_mat<__mpz_struct [1]>&, fplll::ZZ_mat<__mpz_struct [1]>&, fplll::ZZ_mat<__mpz_struct [1]>&, fplll::LLLMethod, int, double, double) ()
   from /lib/x86_64-linux-gnu/libfplll.so.7
#8  0x00007ffff77e6197 in fplll::Wrapper::lll() ()
   from /lib/x86_64-linux-gnu/libfplll.so.7
#9  0x00007ffff77e62fb in int fplll::lll_reduction_wrapper<__mpz_struct [1]>(fplll::ZZ_mat<__mpz_struct [1]>&, fplll::ZZ_mat<__mpz_struct [1]>&, fplll::ZZ_mat<__mpz_struct [1]>&, double, double, fplll::FloatType, int, int) ()
   from /lib/x86_64-linux-gnu/libfplll.so.7
#10 0x00007ffff77e64bb in fplll::lll_reduction(fplll::ZZ_mat<__mpz_struct [1]>&, double, double, fplll::LLLMethod, fplll::FloatType, int, int) ()
   from /lib/x86_64-linux-gnu/libfplll.so.7
#11 0x000055555555e258 in LLLReduce (B=B@entry=0x7fff74001680, d=d@entry=6,
    n=n@entry=9, precision=<optimized out>) at fplll_wrapper.cpp:165
#12 0x000055555555bba2 in slz (Res=Res@entry=0x7ffff4b48630,
    Nb=Nb@entry=0x7ffff4b4862c, P=P@entry=0x7fff740010a0,
    C=C@entry=0x7ffff4b48650, T=T@entry=0x7ffff4b48680, d=d@entry=2, alpha=2,
    tmp=0x7ffff4b48740, W=0x7ffff4b48710, npols=6, ncoefs=9,
    ztmp=<optimized out>, Calpha=0x7ffff4b486c0, ztmp2=0x7ffff4b486b0,
    ok=0x7fff74000f90, PP=0x7fff74000fc0, Ptmp=0x7fff74001000,
    B_lll=0x7fff74001680, time_4_lll=0x7ffff4b48638,
    time_4_Hensel=0x7ffff4b48640, Sol=0x7ffff4b48720, h=0x7ffff4b488c0)
    at slz.c:407
```

joerowell:
This PR fixes #479 by replacing the statically instance of ```mpfr_t``` with a temporary on each call to the functions in ```LDConvHelper```. This is a definite performance regression, but I have no idea if these functions are used much.

### DIFF
--- a/fplll/lll.h
+++ b/fplll/lll.h
@@ -36,7 +36,9 @@ public:
    * needed).
    */
   LLLReduction(MatGSOInterface<ZT, FT> &m, double delta, double eta, int flags);
-
+#ifdef FPLLL_WITH_LONG_DOUBLE
+  ~LLLReduction() { LDConvHelper::free(); }
+#endif
   /**
      @brief LLL reduction.
 
--- a/fplll/lll.h
+++ b/fplll/lll.h
@@ -37,10 +37,6 @@ public:
    */
   LLLReduction(MatGSOInterface<ZT, FT> &m, double delta, double eta, int flags);
 
-#ifdef FPLLL_WITH_LONG_DOUBLE
-  ~LLLReduction() { LDConvHelper::free(); }
-#endif
-
   /**
      @brief LLL reduction.
 
--- a/fplll/nr/nr_Z.inl
+++ b/fplll/nr/nr_Z.inl
@@ -236,19 +236,26 @@ public:
  */
 #ifdef FPLLL_WITH_LONG_DOUBLE
 
+// Thread_local variables for doing numeric conversions.
+// These are global to replace their previous usage, but member variables
+// cannot be thread local and so they're declared at this scope.
+// NOTE: these are extern because these can only have one definition.
+// These are declared in util.cpp
+extern thread_local mpfr_t temp_mpfr;
+extern thread_local bool temp_mpfr_initialized;
+
 class LDConvHelper
 {
 public:
   /** Converts op to a long double with rounding to nearest. */
   static long double mpz_get_ld(const mpz_t op)
   {
-    mpfr_t temp;
-    mpfr_init2(temp, numeric_limits<long double>::digits);
-    mpfr_set_z(temp, op, GMP_RNDN);
-    const auto result{mpfr_get_ld(temp, GMP_RNDN)};  // exact
-    mpfr_clear(temp);
-    return result;
+    init_temp();
+    mpfr_set_z(temp_mpfr, op, GMP_RNDN);
+    return mpfr_get_ld(temp_mpfr, GMP_RNDN);  // exact
   }
+
+  static void free() { free_temp(); }
 
   /**
    * Returns d and sets exp such that 0.5 <= |d| < 1 and d * 2^exp is equal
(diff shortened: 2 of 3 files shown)


cr-marcstevens:
What was exactly the problem here?
That the temp variable was created and freed multiple times, or is even access by multiple threads an issue?

joerowell:
Both, I think. I think there's (at least) two failure conditions in the old version:

- Two threads modify the same temporary. This will give incorrect results, but probably not cause a crash.
- The old ``is_initialized`` flag isn't atomic, and so the check on ``temp_initialized`` isn't guaranteed to do the right thing. In that case we get the double-free observed in #479. 

If this causes a noticeable slow-down, we could always use an object pool instead. This seemed simpler as a first approximation, though.

zimmermann6:
with my application, I see no slowdown. On a 32-core machine, with 32 threads, the wall clock time is 5min19s with the master branch, and 5min14s with this branch. With hyperthreading (64 threads) it is respectively 4min33s and 4min32s.

cr-marcstevens:
@malb Is there a particular use case that could be impacted most? maybe the `proved` case? As a regression test?

The commit looks fine. It just replaces global static variable with in-function temporaries.

BTW There might be another way to be thread safe and avoid initializing/freeing temporaries for each function call. Since we use C++11 anyway, one can also replace global `static` variable with a global `thread_local` variable.

joerowell:
@cr-marcstevens I deliberately did not do the ``thread_local`` approach because I (for some reason) thought that it did not work for OpenMP or pthreads (i.e I thought it only worked for C++11 threads). 

However, having just tested that, you're right: it does appear to be the better solution. I'll change this patch to have that instead (since it'll almost certainly be better if we do find a regression later). Thanks!

zimmermann6:
once the patch is updated, I can test it again on my application

zimmermann6:
I check again with the new version of the branch, on my application (bacsel revision ce2ee73):
```
$ time ./bacsel -rnd_mode all -prec 128 -n 53 -nn 53 -m 43 -t 19.6 -t0 4503599627370496 -t1 5629499534213120 -d 2 -alpha 2 -v -e_in -7 -nthreads 64
```
With master I get 43.5 seconds (45m59s wall clock), and with the new version of the branch I get 43.5s too (46m9s). Thus no slowdown on my side. Thank you!

PR #484, opened 2022-02-02
Use CVPM_PROVED for CVP

ludopulles:
The CVP call from the commandline was missing the CVP-method parameter in the call.
Also, the flags were treated the same for SVP as for CVP, I splitted that into two.


malb:
Ta!

PR #491, opened 2022-06-02
configure.ac: fix bashism in --with-qd test.

orlitzky:
There's a double-equals test here that doesn't work properly when `/bin/sh` is not bash. We simply change it to single-equals.


malb:
Ta!

PR #492, opened 2022-07-16
high-level babai method for GSO objects

malb:


fplll/gso.cpp
```diff
@@ -402,6 +402,112 @@ template <class ZT, class FT> void MatGSO<ZT, FT>::size_increased()
...
+
+template <class ZT, class FT>
+void MatGSO<ZT, FT>::babai(vector<ZT> &w, const vector<FT> &v, int start, int dimension, bool gso)
+{
+  dimension = (dimension == -1) ? this->d - start : dimension;
+  vector<FT> x;
+  if (!gso)
+  {
+    from_canonical(x, v, start, dimension);
+    MatGSOInterface<ZT, FT>::babai(w, x, start, dimension);
```
joerowell, line 503: Should this be ```MatGSOInterface``` here? I'm wondering if you'd ever want to do this rather than explicitly calling ```MatGSO```'s babai's directly. 

To be clear: this would call ```MatGSOInterface<ZT, FT>```'s babai, which would then virtually dispatch to ```MatGSO<ZT, FT>```'s babai. A smart compiler might even de-virtualise this. So there's no correctness issue here: perhaps just a clearness one.

fplll/gso.cpp
```diff
@@ -402,6 +402,112 @@ template <class ZT, class FT> void MatGSO<ZT, FT>::size_increased()
...
+  }
+  else
+  {
+    MatGSOInterface<ZT, FT>::babai(w, v, start, dimension);
```
joerowell, line 507: Same here, I think.

joerowell (approved):
LGTM, just two comments / questions, but feel free to ignore.

fplll/gso.cpp
malb: I was going with "communicating intent" a la: To do the GSO basis transform, we run it using our basis and then call the generic superclass code on the transformed input. Perhaps those methods shouldn't be virtual though?

joerowell: Ah good point. I'm ambivalent on keeping them virtual: whilst we don't need it now, it does communicate intent well.

PR #498, opened 2022-09-30
Compute denominator directly in slope calculation

ludopulles:
This provides a (hopefully) small speed up to calculation of the current slope, although this is probably negligible in the bigger picture.

- Remove any memory usage in the vector x as we can compute the numerator directly.
- The denominator is now calculated with: $$\sum^{n-1}_{i=0} \left(i - \frac{n-1}{2} \right)^2 = \frac{(n+1)n(n-1)}{12}.$$
- Change the documentation to say that the curve is a line fitted with least squares method.

### DIFF
--- a/fplll/gso_interface.cpp
+++ b/fplll/gso_interface.cpp
@@ -199,26 +199,20 @@ double MatGSOInterface<ZT, FT>::get_current_slope(int start_row, int stop_row)
 {
   FT f, log_f;
   long expo;
-  vector<double> x;
-  x.resize(stop_row);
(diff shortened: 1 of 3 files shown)


malb:
Cool! Do we have any tests that would test if those two implementations return the same output (up to numerical noise)?

ludopulles:
I got finally to it to write some tests in commit 4bc7404ca279a19cd83300f1d1a8cb3ce94e5c11.

malb:
Looks good to me

malb:
Thanks!

Issue #500, opened 2022-12-25
test_babai fails on all architectures but amd64


txgk:
When upgrading fplll package in Alpine Linux repositories to 5.4.3 version I ran into a failing test `test_babai` which successfully runs only on amd64 (x86_64).

Upgrade merge request: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/42620
aarch64 build log: https://gitlab.alpinelinux.org/grikir02/aports/-/jobs/934088
armhf build log: https://gitlab.alpinelinux.org/grikir02/aports/-/jobs/934090
armv7 build log: https://gitlab.alpinelinux.org/grikir02/aports/-/jobs/934089
ppc64le build log: https://gitlab.alpinelinux.org/grikir02/aports/-/jobs/934087
s390x build log: https://gitlab.alpinelinux.org/grikir02/aports/-/jobs/934086
x86 build log: https://gitlab.alpinelinux.org/grikir02/aports/-/jobs/934085

Can you provide a patch which fixes this regression?

SnarkBoojum:
I confirm the issue on Debian; you can access the logs [here](https://buildd.debian.org/status/package.php?p=fplll).

malb:
Thanks for reporting these. This should be now fixed in master. I assume for you to test it you'd prefer a new release cut?

txgk:
If that's okay with you, it would be cool. Bumping a version is easier than applying patch :)

SnarkBoojum:
I uploaded 5.4.3-2 which should appear in the logs in a few hours, with just the simple commit as a patch.

I suggest to wait until the fix is confirmed before cutting a new version.

joerowell:
Thanks -- I only tested the patch on ARM and x86, so waiting for the other
platforms is a good idea in case I missed something obvious.

musicinmybrain:
I’m the primary maintainer of the [Fedora Linux package](https://src.fedoraproject.org/rpms/libfplll). I’m currently verifying the patch on x86_64, i686, ppc64le, s390x, aarch64, and armhfp/armv7hl. I’ll follow up when the test builds finish.

Can you please have a look at https://github.com/fplll/fplll/issues/502 before cutting a new release?

joerowell:
I think the patch needs an update for ppc etc -- we should also cover the
case that long double is 128 bits, which I forgot earlier.

I'll do that tomorrow, and hopefully that'll fix all of the builds.

musicinmybrain:
[...]

PR #501, opened 2022-12-29
Fix failing babai test.

joerowell:
TL;DR: ```long double``` isn't always 80 bits in length.

The old tests implicitly assumed that ```long double``` was more precise than ```double```:

https://github.com/fplll/fplll/blob/322b0a1c7db191a134ab3f7ab8af4597471b0af8/tests/test_babai.cpp#L86-L88

[However, this isn't true on all platforms](https://en.wikipedia.org/wiki/Long_double). Indeed, on my M1 Mac, both ```long double``` and ```double``` are 64-bits. 

This PR makes the behaviour of the test conditional on this detail. I also added some more printing, because it's easier to tell why the test is failing with more info :) 

This should address #500. Maybe we should add other platforms to the CI too?

### DIFF
--- a/tests/test_babai.cpp
+++ b/tests/test_babai.cpp
@@ -68,8 +68,9 @@ int test_intrel(long n, long bits, bool shouldfail = false, int seed = 0)
   }
   else
   {
-    std::cerr << "n:" << n << ", bits: " << bits << std::endl;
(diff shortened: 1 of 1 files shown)


malb:
Thank you!!

PR #503, opened 2022-12-30
Trivial typo fix in README.md

musicinmybrain:
Replaces `contructed` with `constructed`.


musicinmybrain:
I don’t think this was intentional, although it could be a medium-quality pun on “coNTRUcted.”

malb:
Something when wrong with the merge, PR is merged but still open thus simply closing it now. Thanks!

PR #504, opened 2022-12-30
Make Babai test more robust for PPC etc.

joerowell:
Turns out I missed the case for ```long double``` being synonymous with ```__float128``` and a handful of other cases. This PR should finish #500.

### DIFF
--- a/tests/test_babai.cpp
+++ b/tests/test_babai.cpp
@@ -16,6 +16,8 @@
 #include "fplll/fplll.h"
 #include <vector>
 
+#include<cfloat> // Needed for precision macros. 
+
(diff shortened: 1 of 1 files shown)


joerowell:
@malb Should we also try the commented out tests for QD? 034e8cb might make those tests more reliable too.

malb:
Oh, cool, yes that'd be great.

typed with thumbs, please excuse my brevity

joerowell:
Sadly it doesn't fix the QD tests: we can fix that in a later PR?

malb:
Yeah, let's leave that for another time!

PR #511, opened 2023-02-23
Remove FPLLL_SONUM from fplll/fplll_config.h.in

### LINKED ISSUE #509: FPLLL_SONUM macro has no substitution
In `fplll/fplll_config.h.in`, the `FPLLL_SONUM` preprocessor macro is defined to have the value of the `@FPLLL_SONUM@`:

https://github.com/fplll/fplll/blob/3e10c51b31f29fca4a6b1e11494b96a5009dd19d/fplll/fplll_config.h.in#L16-L17

Since there is no `AC_SUBST([FPLLL_SONUM], ...)`, nothing is substituted here, and the template `@FPLLL_SONUM@` appears in the generated API header `fplll/fplll_config.h`.

Either this definition should be removed from the header, or an appropriate value should be defined with `AC_SUBST`. If you know what you want to do about this, I am happy to offer a PR.

This was originally reported [downstream in Fedora Linux](https://bugzilla.redhat.com/show_bug.cgi?id=2172617).

musicinmybrain:
Nothing was ever substituted for the template `@FPLLL_SONUM@`, so the template was appearing verbatim in the generated `fplll/fplll_config.h`. Since that isn’t a usable value, we know that nobody was relying on this particular macro as part of the API.

Fixes #509.

### DIFF
--- a/fplll/fplll_config.h.in
+++ b/fplll/fplll_config.h.in
@@ -13,9 +13,6 @@
 /* fplll micro version */
 #define FPLLL_MICRO_VERSION @FPLLL_MICRO_VERSION@
 
-/* fplll sonum */
-#define FPLLL_SONUM @FPLLL_SONUM@
-
 /* fplll version */
 #define FPLLL_VERSION @FPLLL_VERSION@
 


musicinmybrain:
I’d much rather remove this macro than try to replicate libtool library versioning math in a way that doesn’t surprise anyone. It’s obviously not something that anyone has been clamoring for, since it’s been broken and nobody has noticed until now.

malb:
Thanks!

PR #512, opened 2023-03-04
Improve default strategy path handling

### LINKED ISSUE #510: Handling FPLLL_DEFAULT_STRATEGY(_PATH) for system-wide library
I’m the current primary maintainer of the [`libfplll` package in Fedora Linux](https://src.fedoraproject.org/rpms/libfplll). I just learned from a [downstream bug report](https://bugzilla.redhat.com/show_bug.cgi?id=2172617) and from https://github.com/fplll/fpylll/issues/221 that the `FPLLL_DEFAULT_STRATEGY` and `FPLLL_DEFAULT_STRATEGY_PATH` macros in the installed API headers don’t have useful values (both are `""`).

It looks like there is no provision to define those properly when building packages; the empty-string fallback values are defined in `fplll/defs.h`, and the macros aren’t otherwise defined in any headers.

https://github.com/fplll/fplll/blob/3e10c51b31f29fca4a6b1e11494b96a5009dd19d/fplll/defs.h#L283-L289

The reason that the experiment in https://github.com/fplll/fpylll/issues/221#issuecomment-1031488375 yielded appropriate 

These macros are defined using compiler flags while building the library, so the *compiled library* does have the correct values available in [`fplll/bkz_param.cpp`](https://github.com/fplll/fplll/blob/5.4.4/fplll/bkz_param.cpp), but API users can’t access them without calling a function to see what values were compiled into the library.

https://github.com/fplll/fplll/blob/3e10c51b31f29fca4a6b1e11494b96a5009dd19d/fplll/Makefile.am#L5

It seems like the default values used at build time should be formed in `configure.ac` and `AC_SUBST`ed into [`fplll/fplll_config.h.in`](https://github.com/fplll/fplll/blob/5.4.4/fplll/fplll_config.h.in), which produces `fplll/fplll_config.h`, which is an API header and is `#include`d from `fplll/defs.h`. That would make the defaults used at compile time available via the API headers, just as is already the case for some compile-time values like `FPLLL_MAX_ENUM_DIM`.

What do you think? Is this how you think these macros should work? I’m happy to provide a PR.

musicinmybrain:
Do not define `FPLLL_DEFAULT_STRATEGY_PATH` or `FPLLL_DEFAULT_STRATEGY` to the empty string (when undefined) in fplll/defs.h. This is not usable for library API users, and not needed for compiling the library.

Allow setting the default strategy paths via new `configure` script options `--with-default-strategy-dir` and `--with-default-strategy`. This allows distributions that need the data directory to be in a different place to avoid patching the build system. For example, Debian might want to use `--with-default-strategy-dir=/usr/share/libfplll5/strategies`.

Make the actual strategy paths compiled into the library available as variables in the `pkg-config` file `fplll.pc`.

Fixes #510.

-----

I have tested this by rebuilding all Fedora Linux packages that depend on `libfplll` (`gap-pkg-float`, `linbox`, `Macaulay2`, `python-fplll`, and `sagemath`) [in a COPR](https://copr.fedorainfracloud.org/coprs/music/libfplll/packages/).

I’m hoping that people who commented in https://github.com/fplll/fpylll/issues/221 will review this and verify it meets their needs, or suggest any necessary improvements.


SnarkBoojum:
I like the idea of the configure flags.

PR #516, opened 2023-09-09
std::result_of has been removed in C++20

vbeffara:
`std::result_of` has been deprecated in C++17, removed in C++20 and support has been removed from (some) recent compilers, in particular fplll 5.4.4 does not work with clang 16. This patch replaces `std::result_of` with uses of `decltype`.

### DIFF
--- a/fplll/io/thread_pool.hpp
+++ b/fplll/io/thread_pool.hpp
@@ -96,7 +96,7 @@ namespace thread_pool {
 
 		// enqueue a function and obtain a future on its return value
 		template<typename F, typename... Args>
-		auto enqueue(F&& f, Args&&... args) -> std::future<std::invoke_result_t<F, Args...>>;
+		auto enqueue(F&& f, Args&&... args) -> std::future<decltype(f(args...))>;
 
 		// push a trivial function without a future
 		void push(const std::function<void()>& f);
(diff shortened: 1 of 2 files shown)


vbeffara:
(Sorry about the quick update, the initial code with `invoke_result_t` passed the tests on my config but not on CI ...)

joerowell:
@malb this PR is functionally the same as we have now, so I'm happy with merging it. 

For future reference / use, I have a few notes:

1. We can't use ```decltype``` with pointers to member functions. This isn't a real issue, because we can just wrap the pointers to member functions in lambdas (similarly to what G6K does [here](https://github.com/fplll/g6k/blob/959fd8f290c4c36fc60995b6c1b5f527f94fefc4/kernel/bgj1_sieve.cpp#L114)). 
2. We should probably update G6K too, since the thread-pool there also uses ```std::result_of```. 

I'm happy to defer to @cr-marcstevens, though, since he knows this code far better than I do.

joerowell (approved):

malb:
Thanks! I'll give it a bit of time to see if @cr-marcstevens wants to weight in but merge it in a few days if he's busy otherwise.

vbeffara:
Indeed `invoke_result` is the "official" replacement for `result_of`, your branch works for me (clang 16 on macOS).

cr-marcstevens:
Okay great! This code is in several repos, so I can update those now as well.

@malb Would it be an option to add clang 16 (macos a/o linux) to github actions and see if that would have caught the bug?
(I also noticed there is a actions warning on node 12 vs node 16, which may be due to deprecated action step versions like checkoutv2)

PR #519, opened 2024-03-17
disable long doubles on aarch64-darwin

Feyorsh:
`aarch64-darwin` does not have support for long doubles; please see the corresponding [issue](https://github.com/fplll/fpylll/issues/269) and [fix](https://github.com/fplll/fpylll/pull/270) in `fpylll` for more information.


malb:
Thank you!

PR #520, opened 2024-03-26
Add example code in gso.h

rohan-kalra904:
I have added a small example in the gso.h file which is code but enclosed in comments so, when compiled using doxygen it shows as an example along with the other documentation.
We have discussed it here: 
https://groups.google.com/g/fplll-devel/c/WHFnzVENsN0

### DIFF
--- a/fplll/.vscode/settings.json
+++ b/fplll/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
--- a/fplll/gso.h
+++ b/fplll/gso.h
@@ -108,6 +108,35 @@ public:
    * @param row_op_force_long
    *   Affects the behaviour of row_addmul(_we).
    *   See the documentation of row_addmul.
+   * @section Example Example Usage:
+   *    <code>
+   *    #include <cstring>                                      <br>
+   *    #include <fplll/gso.h>                                  <br>
+   *    #include <fplll/gso_gram.h>                             <br>
+   *    #include <fplll/gso_interface.h>                        <br>
+   *    #include <fplll/householder.h>                          <br>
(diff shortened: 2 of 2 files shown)


malb:
Hi! Thanks, but without automatically testing those examples, I don't think this will work.

PS: Let's not add your VSCode config :)

rohan-kalra904:
Yes, with this way there is no way to automatically test the examples but there is another way by which we can create examples in other .cpp files and then using a doxygen command include it with the relevant documentation.
For this we will have to create an example file for every function or group of similar functions. 
Maybe we can work out a way to test those files.

malb:
Hi there, I'd insist on that "code that isn't tested is broken" so the only option of those two would be those cpp files (?)

rohan-kalra904:
Hi,
I wanted to ask, for testing the cpp files is it enough if they compile and run without errors or should we have an expected output that we compare with the output after running the file? For the 2nd method we can't generate the lattices randomly and would need to decide on an input.

malb:
I'd say "ideally" yes, but we currently only do lightweight checks in our tests, anyway. So I'd say we don't need to do more here than in, say: https://github.com/fplll/fplll/blob/master/tests/test_lll.cpp

Does that work?

rohan-kalra904:
Yes, That will work. I will create a new folder containing the cpp files, each file will first have the example and then a function which tests if the code is still working and this test function will not show in the documentation. But should I create a different example file for each function or can I club some functions together and put them in one file?  Also can you please tell what should I name the folder containing all the examples in it?
Should I close this branch and start working on a new branch altogether, or should I just continue working on this branch?

malb:
Sounds good, I think it makes sense to collect several functions in one file. Perhaps `doctests` is a good directory name? I have no strong opinions about whether changing the PR or not is the right call here :)

Issue #495, opened 2022-08-13
Parallel enumerator and recursive enumerator disagree on node counts


joerowell:
Consider the following test program (placed in ```tests```):

```c++
#include <cstring>
#include <fplll.h>
#include <test_utils.h>

template <typename FT> int test_counters_agree(const size_t d)
{

  RandGen::init_with_seed(0x1337);
  ZZ_mat<mpz_t> A = ZZ_mat<mpz_t>(100, 100);
  A.gen_qary_withq(50, 7681);
  lll_reduction(A);

  ZZ_mat<mpz_t> U;
  MatGSO<Z_NR<mpz_t>, FP_NR<FT>> M(A, U, U, 0);
  M.update_gso();

  FastEvaluator<FP_NR<FT>> evaluator;
  Enumeration<Z_NR<mpz_t>, FP_NR<FT>> enum_obj(M, evaluator);
  FP_NR<FT> max_dist;
  M.get_r(max_dist, 0, 0);
  max_dist *= 0.99;

  // Make sure we get the external enumerator for later.
  // This is because the call to set_external_enumerator
  // clears the global extenum pointer: we need to make sure that we
  // "undo" the change so we can repeat this test.
  const auto ext_enumerator = get_external_enumerator();

  Enumeration<Z_NR<mpz_t>, FP_NR<FT>> enumobj(M, evaluator);

  // N.B the call here changes max_dist: for a fair comparison,
  // we reserve the old distance.
  const auto old_max_dist = max_dist;
  // The first call here uses the parallel enumerator.
  enumobj.enumerate(0, d, max_dist, 0);

  // Check how many nodes the parallel enumeration visited.
  auto par_enum_nodes = enumobj.get_nodes();

  // Now we'll use the regular enumerator.
  set_external_enumerator(nullptr);

  // Go back to using the old distance
  max_dist = old_max_dist;
  enumobj.enumerate(0, d, max_dist, 0);

  // Now check how many nodes the recursive enumeration visited.
  auto rec_enum_nodes = enumobj.get_nodes();

  // Switch back over to the external enumerator
  set_external_enumerator(ext_enumerator);

  std::cerr << "par_enum_nodes for " << d << " = " << par_enum_nodes << std::endl;
  std::cerr << "rec_enum_nodes for " << d << " = " << rec_enum_nodes << std::endl;
  return 0;
}

int main()
{

  int status = 0;
  status |= test_counters_agree<mpfr_t>(30);
[...]

PR #526, opened 2024-06-11
Change choice of q in qary-lattice generation

ludopulles:
This pull request might be a bit controversial, as it contains multiple small things that boil down to personal preference. So perhaps, you want to cherry-pick this pull request, or ask for changes.

## Current situation

Currently, the value for q in the generation of a q-ary lattice could be the same for different parameters, e.g.:

```
$ latticegen q 1 1 7 p
[[127]]
$ latticegen q 1 1 8 p
[[127]]
```
and
```
$ latticegen q 1 1 7 b
[[116]]
$ latticegen q 1 1 8 b
[[116]]
```
show two (similar) cases where the value for `q` is the same, while the argument `<b>`, controlling the number of bits, is different (i.e. 7 and 8 respectively). Note that `b=16, b=17, b=18` all give the same value for q!

Moreover, it may well be that choosing some large value for `<b>` still results in a small-ish value for `q`, e.g.:
```sh
$ latticegen q 1 1 18 b
[[43124]]
```
is of the order `2^{15.40}`, much smaller than `2^18`.

## Proposed changes

1. This pull request mandates that when calling `latticegen` with arguments `q <d> <k> <b> <c>` having <c> equal to `p` or `b`, the value for `q` must be in the interval `[2^{b-1}, 2^b - 1]`, whereas previously the value was in `[0, 2^b-1]`.
Note: for `p`, actually the eventual value could be larger if the next prime was further on.
2. The internal methods `gen_qary(int, int)` and `gen_qary_withq(int, int)` are changed to `gen-qary_bits(int, int)` and `gen_qary(int, int)` respectively in the file `fplll/nr/matrix.h`, because the function `gen_qary(int, Z_NR<ZT>)` also expects a value for `q` in the second argument, so this naming makes more sense.
3. Sort .gitignore, remove `Makefile` & `Makefile.in` (these occured 2x) and `.gitignore` (seems to be ignored by .gitignore?), and add `/tests/test_ceil`.

## Breaking changes
Line 10 in the sample code of #495 would have to change from `A.gen_qary_withq(50, 7681);` to `A.gen_qary(50, 7681);`.


ludopulles:
Odd, `clang-format` seems to give a different result with CI compared to my locally compiled (version `clang-format version 19.0.0git (https://github.com/llvm/llvm-project.git 149ed9d2c58095f87745f8696ef96b5076c91fca)`)

malb:
I think different versions of clang have different style demands, I tend to just obey the CI :)

ludopulles:
OK, I updated the formatting just now.

m4/ax_cxx_compile_stdcxx.m4
malb, line 1: Why does this file change?

m4/ax_pthread.m4
malb, line 1: why does this file change?

cr-marcstevens: Probably because he ran the M4 update script in the fplll repo: .check-m4.sh

malb: ah, these should be taken out of this PR, I'd say :)

ludopulles: Indeed, the `make check-style` command (see [CI](https://github.com/fplll/fplll/actions/runs/9497128849/job/26173009017?pr=526)) indicated there were 2 new updates available:
```
Checking for newer version of: m4/ax_cxx_compile_stdcxx.m4
*** UPDATE AVAILABLE for m4/ax_cxx_compile_stdcxx.m4: version 12 (current) => 18 (latest) ***
Checking for newer version of: m4/ax_pthread.m4
*** UPDATE AVAILABLE for m4/ax_pthread.m4: version 30 (current) => 31 (latest) ***
```
I have removed it from the PR.

malb:
Thanks! Could you also submit a PR to FPyLLL to update it to  the API change and also check if G6K still works with both PRs applied? Sorry, I know I'm asking a lot, but we may have gotten to the point in the life of FPLLL where we need to be a bit more careful about API changes like this (?)

ludopulles:
Note: I noticed that `latticegen` and IntegerMatrix.random(q=[...]) didn't support values of 'q' bigger than INT_MAX, so this PR also supports those by casting the command line string directly to mpz_t.

malb:
Looks good to me.

PR #528, opened 2024-06-16
Unify header include style 

### LINKED ISSUE #527: Mixed style of Header inclusion
I'm in the process of migrating build from libtool to cmake, and got bit confused with mixed styles of header inclusion.

Sometimes we see relative path "../defs.h", sometimes library root path "fplll/defs.h", sometimes system path <fplll/defs.h>

And is the assumption that some header files are installed into {prefix}/include first before I can build libfplll an important one? 

Would like to get clarifications on the reason behind this. Or is it just legacy issue and could all be refactored to using relative path?

alxiong:
Close #527 

As discussed, I unify all import style under `fplll/*` to relative path, and all under `tests/` to be system path (since tests normally assume built package and installed headers) 

~Also the `cmake` build config is only my best effort, and not super well tested. I locally run, the artifacts are as close to the `automake` tool as possible. Hopefully this is just pure additional and completely optional to the developers to use it.  (added the "unofficial" warning in README)~

~also happy to drop [12b49aa](https://github.com/fplll/fplll/pull/528/commits/12b49aaf25766369932aeeb01d469a8f786b7aaa) if it's undesirable~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,7 @@ stamp-h1
 /tests/test_sieve
 /tests/test_svp
 /tests/test_svp_gram
+**/*.DS_Store
+config.guess~
+config.sub~
+install-sh~
--- a/fplll/Makefile.am
+++ b/fplll/Makefile.am
@@ -7,9 +7,7 @@ EXTRA_DIST = io/json.hpp ballvol.const factorial.const
 
 CLEANFILES = *.gcov .libs/*.gcda .libs/*.gcno *.gcno *.gcda */*.gcno */*.gcda
 
-nodist_include_fplll_HEADERS=fplll_config.h
-
-nobase_include_fplll_HEADERS=defs.h fplll.h \
+nobase_include_fplll_HEADERS=defs.h fplll.h fplll_config.h \
 	nr/dpe.h \
 	nr/matrix.h \
 	nr/matrix.cpp \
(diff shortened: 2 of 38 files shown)


malb:
Thank you *so* much tackling such a intense maintenance task!

1. Can separate our the cmake contribution from the headers contribution, the former will be a different discussion than the latter.
2. Did you test also if FPyLLL/SageMath still work with the new headers?

alxiong:
> Can separate our the cmake contribution from the headers contribution

sure, will submit a separate PR

> 2. Did you test also if FPyLLL/SageMath still work with the new headers?

not yet, I'll do that next step, need to fix CI failure first.

will ping you again when ready for review again.

malb:
Thanks! Much appreciated!

alxiong:
@malb the previously failed CI make commands all pass on locally, and I rebuild `python setup.py build` for fpylll with the latest `$VIRUTAL_ENV/lib` without issues. 
(i didn't build sage, but i wouldn't imagine a problem there)

so, ready for review and merge, imo.

malb:
This looks good to me, @cr-marcstevens or @joerowell do you have strong opinions about this? If not, I'd merge.

joerowell:
I'm for this PR (indeed, thanks @alxiong!), but I'd recommend we put some thought into how we're going to protect against drift here. This was a lot of manual effort, and I'd hate for it to have to be repeated again. 

Moreover, as far as I'm aware, this PR doesn't update the clang-format file to reject certain include paths. I think clang-format could be configured to do that, but I'm not sure.

malb:
I'd be okay with merging this and worrying about the points of @joerowell in a separate PR? This is mostly because this PR has a high risk of bitrot?

PR #531, opened 2024-06-18
Remove warning for >=C++20 compilers

ludopulles:
There was the following warning when running `make`, when compiling the C++ code with a modern compiler (i.e. one that supports C++20 but still compiles with C++11):

	warning: template-id not allowed for constructor in C++20 [-Wtemplate-id-cdtor]
	note: remove the '< >'

More info, see: https://stackoverflow.com/questions/63513984/can-class-template-constructors-have-a-redundant-template-parameter-list-in-c2

### DIFF
--- a/fplll/nr/nr_FP.inl
+++ b/fplll/nr/nr_FP.inl
@@ -27,11 +27,11 @@ public:
   /**
    * Constructors.
    */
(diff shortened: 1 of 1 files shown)


malb:
Ta!

PR #537, opened 2025-04-28
Update LPGL 2.1 license text; update LGPL-2.1-or-later license notice

musicinmybrain:
The FSF is now remote-only and no longer has a physical address.

Updated `COPYING` from https://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt, and updated the header of `fplll/nr/dpe.h` based on https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html#SEC4.


malb:
Thanks, obv these failures aren't the fault of this PR, will need to look into that.

PR #552, opened 2026-07-29
Fix build on MinGW

Suyun114:
Closes #542

### DIFF
--- a/fplll/defs.h
+++ b/fplll/defs.h
@@ -23,14 +23,19 @@
 #define FPLLL_WITH_LONG_DOUBLE
 #endif
 
+#if !defined(__MINGW32__) && !defined(_MSC_VER)
+#define FPLLL_WITH_GETRUSAGE
+#endif
+
 #define FPLLL_WITH_DPE
 #define FPLLL_WITH_ZDOUBLE
 #define FPLLL_WITH_ZLONG
-#define FPLLL_WITH_GETRUSAGE
+
+#define _USE_MATH_DEFINES
+#include <cmath>
 
 #include <algorithm>
 #include <climits>
-#include <cmath>
 #include <cstddef>
 #include <cstdio>
 #include <cstdlib>
@@ -42,10 +47,10 @@
 #include <sstream>
 #include <stdexcept>
 #include <string>
+#include <sys/time.h>
 
 #ifdef FPLLL_WITH_GETRUSAGE
 #include <sys/resource.h>
-#include <sys/time.h>
 #include <unistd.h>
 #endif
 
--- a/fplll/nr/nr_Z.inl
+++ b/fplll/nr/nr_Z.inl
@@ -237,14 +237,6 @@ public:
  */
 #ifdef FPLLL_WITH_LONG_DOUBLE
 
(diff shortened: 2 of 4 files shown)


malb:
We won't accept Windows in the CI because we cannot support it, no maintainer runs windows. We can accept fixes that are aimed at Windows on a "best effort" basis as long as they don't break on Linux and OSX.

Suyun114:
> We won't accept Windows in the CI because we cannot support it, no maintainer runs windows. We can accept fixes that are aimed at Windows on a "best effort" basis as long as they don't break on Linux and OSX.

The CI part has been removed now.

fplll/nr/nr_Z.inl
```diff
@@ -237,23 +237,15 @@ public:
...
 #ifdef FPLLL_WITH_LONG_DOUBLE
 
-// Thread_local variables for doing numeric conversions.
-// These are global to replace their previous usage, but member variables
-// cannot be thread local and so they're declared at this scope.
-// NOTE: these are extern because these can only have one definition.
-// These are defined in util.cpp
-extern thread_local mpfr_t temp_mpfr;
-extern thread_local bool temp_mpfr_initialized;
-
```
malb, line 247: Why this change?

fplll/defs.h
```diff
@@ -23,14 +23,19 @@
 #define FPLLL_WITH_LONG_DOUBLE
 #endif
 
+#if !defined(__MINGW32__) && !defined(_MSC_VER)
```
malb, line 26: Why this change?

fplll/defs.h
```diff
@@ -23,14 +23,19 @@
...
+#define FPLLL_WITH_GETRUSAGE
+#endif
+
 #define FPLLL_WITH_DPE
 #define FPLLL_WITH_ZDOUBLE
 #define FPLLL_WITH_ZLONG
-#define FPLLL_WITH_GETRUSAGE
+
+#define _USE_MATH_DEFINES
```
malb, line 34: Why this change?

fplll/util.cpp
```diff
@@ -27,15 +27,6 @@ enum MinPrecAlgo
   MINPREC_L2
 };
 
-/*
-  Helper variables for LDConvHelper (declared in nr_Z.inl, must be defined in exactly
-  one source file)
```
malb, line 32: why this change?

fplll/nr/nr_Z.inl
Suyun114: Explained on line 273.

fplll/defs.h
Suyun114: Windows does not have `getrusage` syscall.

fplll/util.cpp
Suyun114: Explained on line 273 in `fplll/nr/nr_Z.inl`.

fplll/defs.h
Suyun114: `cmath`'s behavior in Linux is different from Windows. `_USE_MATH_DEFINES` is needed for marcos such as `M_PI` and `M_E` with `-std=c++11` in Windows.

